PR #1, opened 2017-01-26
Move handling of LOGENTRIES_KEY into separate shell script.

forevermatt:


forevermatt:
To make it easier to set up the logentries key in the rsyslog.conf file in projects that use ssp-base, I'd like to split out that logic from the run.sh script. You all okay with that?

PR #7, opened 2017-05-11
Feature/silauth saml attrs

forevermatt:


forevermatt:
Note: This does step back from requiring `"silinternational/ssp-deps": "^1.0"` to "silinternational/ssp-deps": "dev-develop"`, but as long as that makes it back to a tagged version number before we release ssp-base, we should be good. I think this is a necessary step to untangling the rest of our repos from ssp-deps.

PR #9, opened 2017-06-15
Install GMP extension for PHP.

forevermatt:


forevermatt:
This is adding an extension needed by the IP address PHP library I'm using for whitelisting in SilAuth.

PR #17, opened 2017-07-13
Stop specifying version of php-gmp extension.

forevermatt:
This is to match the similar change made in the php7 docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ MAINTAINER Phillip Shipley <phillip.shipley@gmail.com>
 ENV REFRESHED_AT 2016-12-16
 
 RUN apt-get update -y && \
-    apt-get install -y php-memcache php7.0-gmp && \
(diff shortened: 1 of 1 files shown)


fillup:
That extension has been added to silintl/php7 now so it is no longer needed in ssp-base as well, so you can actually remove it from here I believe.

wcjr:
so removing it would simply mean there were never any direct dependencies on `php-gmp` from @Baggerone's code, right?

forevermatt:
There are direct dependencies on `php-gmp`. Including it here in ssp-base isn't _necessary_ at this point, since it's being provided in out php7 image, but it seems better to specify it here as well since ssp-base pulls in code that has that direct dependency on php-gmp (which also protects us if it's ever removed from php7).

wcjr:
<$0.02>I agree, if base depends on it I think it should be listed in the composer file regardless of what the underlying infrastructure may look like.</$0.02>

PR #18, opened 2017-07-14
Material updates, expiry check, etc. 

fillup:


development/idp-local/config/config.php
```diff
@@ -63,6 +83,11 @@
      */
     'tempdir' => '/tmp/simplesaml',
 
+    /*
+     * Name of this IdP to display to the user
+     */
+    'idp_name' => $IDP_NAME,
```
forevermatt: Not that it affects this PR, but here's an example of where the expectation is that `$IDP_NAME` contains a user-friendly version (rather than something like a URL-friendly version).

tests/MetadataTest.php
```diff
@@ -12,7 +12,9 @@
 class MetadataTest extends TestCase
 {
     // SP Metadata entry to request that certain tests be skipped
-    const SkipTestsKey = "SkipTests";    
+    const SkipTestsKey = "SkipTests";
```
forevermatt: It's probably not critical, but it may be worth changing these two constants to use ALL_CAPS rather than TitleCase (per our coding conventions).

tests/MetadataTest.php
```diff
@@ -56,20 +58,90 @@ public function testMetadataFilesReturnArrays()
...
+                self::IdpCode .' element that has something other than letters, ' .
+                'numbers, hyphens and underscores. IDP: ' . $entityId);
+        }
+    }
+
+    public function testIDPRemoteMetadataBadSPList()
+    {
+        $hubMode = Env::get('HUB_MODE', true);
+        if ( ! $hubMode) {
+            $this->markTestSkipped('Skipping test because HUB_MODE = false');
```
forevermatt: Are we somehow running these tests with `HUB_MODE` set to `true` and also with it set to `false`? Having tests that are only conditionally run seems suspicious to me.

Baggerone: These aren't the normal type of tests for an app.  They are for codeship to run to test the contents of the metadata files (and maybe the config?) pre-deployment.  Some of the tests only apply to a hub setup and not to an IdP.

forevermatt: So are the tests being run twice, once with HUB_MODE set to true and once with it set to false?

Baggerone: The tests are run during the build of an implementation instance, not during a build of ssp-base just by itself.  So, when ssp-hub-insite-staging is built by codeship, those tests are run with HUB_MODE set to true focusing on the metadata for that implementation.  I assume that in an idp-in-a-box implementation, HUB_MODE will be set to false and so the tests will run that way for that implementation.

Baggerone: The things that ssp-base depends on have there own tests that are run before they ever get pulled into ssp-base.

forevermatt:
We're waiting to merge this because of the bugs we came across running this in staging, right? If that's the case, it may be worth having one of us submit a rejecting review so that this doesn't look like it's ready to be merged.

fillup:
Closing this PR, will create a new one when upcoming changes are in develop.

PR #24, opened 2017-07-28
Use IDP_DISPLAY_NAME if available.

forevermatt:
This makes the IdP's display name available in the config, which will allow us to use it from the material theme.

Since the material theme is just a theme (which others could use), I don't know if we should have the material theme conditionally look for idp_display_name in the config (and if not present, use the idp_name config instead) or if it can simply expect an idp_display_name to be present.


wcjr:
Do we need to look through all the other modules to make sure they're using it in the right place?

forevermatt:
@wcjr Good question. It might not hurt, but my guess is that we will notice (when using the IdPs) places where things are using the wrong one, so I don't think hunting through the modules needs to be a high priority item. Just my 2 cents.

PR #27, opened 2017-08-30
Release 2.0.0

Baggerone:
Updated dependencies, including a major version of silauth


dockerbuild/ssp-overrides/config.php
```diff
@@ -43,6 +70,31 @@
...
+     'analytics.trackingId' => $ANALYTICS_ID,
+
+    /*
+     * Get a string of html to show as an announcement on the discovery page
+     * and/or login page.  By default, this will be fetched from
+     *   .../vendor/simplesamlphp/simplesamlphp/announcement/announcement.php
+     */
+    'announcement' => AnnouncementUtils::getAnnouncement(),
+    
+    'passwordChangeUrl' => $PASSWORD_CHANGE_URL,
```
fillup: with these vars defined here, do they also need to be in `saml20-idp-hosted.php`? Or did we just put it in there in our appsdev instance for testing?

forevermatt: I don't know. Perhaps @Baggerone does.

Baggerone: This is my current understanding. The idp-hosted metadata includes an authproc call to the expirychecker.  That authproc needs to be told what the passwordChangeUrl is by the idp-hosted.php file.  It is currently doing that using Env::requireEnv('PASSWORD_CHANGE_URL').  

The PASSWORD_FORGOT_URL isn't used there.

PR #47, opened 2018-01-31
"Get new codes here" functionality, plus SSP 1.15

longrunningprocess:
@fillup waiting on decision regarding https://github.com/silinternational/simplesamlphp-module-material/issues/54 before mergin to `master`

### DIFF
--- a/development/idp-local/config/config.php
+++ b/development/idp-local/config/config.php
@@ -173,18 +173,18 @@ $config = [
      * Logging.
      *
      * define the minimum log level to log
-     *		SimpleSAML_Logger::ERR		No statistics, only errors
-     *		SimpleSAML_Logger::WARNING	No statistics, only warnings/errors
-     *		SimpleSAML_Logger::NOTICE	Statistics and errors
-     *		SimpleSAML_Logger::INFO		Verbose logs
-     *		SimpleSAML_Logger::DEBUG	Full debug logs - not reccomended for production
+     *		SimpleSAML\Logger::ERR		No statistics, only errors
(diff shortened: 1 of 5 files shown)


forevermatt:
Just so that we're all aware, we need to wait to merge this until we finish checking (and where necessary, updating) our SimpleSAMLphp modules to work with simplesamlphp 1.15.

forevermatt:
I think we're finally ready to cut a new release (minor version bump, I think) of `ssp-base`. @fillup / @wcjr / @Baggerone, does that sound right to you?

forevermatt:
Okay, this seems to be working fine on the staging hub. I think we're ready to release.

forevermatt:
As far as I can tell, it's time to release and tag this update to ssp-base. @wcjr @Baggerone @fillup Any objections?

Baggerone:
If the new version of simplesamlphp is a breaking change, why wouldn't this be considered one also?

forevermatt:
@Baggerone made some interesting points as we were discussing this:
- Since simplesamlphp is designed to allow using custom modules, and switching from 1.14 to 1.15 does (I think unintentionally) include a backwards breaking change affecting any modules using the old namespacing in `instanceof` or typehints, that seems to suggest this should be a major version bump.
- Is there any reason _not_ to bump the major version? I can't necessarily think of one (since, depending on what you're "measure", this does break some backwards compatibility).

@fillup Thoughts?

wcjr:
I agree, it should be a major then.  You never know if someone might be relying on our base or not...

fillup:
Thats fine, major that is.

PR #56, opened 2018-02-19
get sildisco bug fix

Baggerone:


forevermatt:
Does it matter that this also updates Yii from 2.0.13.1 to 2.0.14?
http://www.yiiframework.com/news/165/yii-2-0-14-is-released/

PR #91, opened 2019-05-22
Update to SimpleSAMLphp 1.17

Schparky:
Several dependencies are needing to be updated as well. Currently using `develop` but will update to a release tag when those are ready.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,13 +37,13 @@
     "roave/security-advisories": "dev-master",
     "simplesamlphp/simplesamlphp": "~1.17.1",
     "simplesamlphp/composer-module-installer": "1.1.6",
-    "silinternational/simplesamlphp-module-expirychecker": "dev-feature/update-to-ssp-1-17",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -10,18 +10,40 @@
     }
   ],
   "minimum-stability": "stable",
+  "repositories": [
+    {
+      "type": "vcs",
+      "url": "git@github.com:silinternational/simplesamlphp-module-expirychecker"
```
forevermatt: What do these entries in `repositories` do?

Schparky: Pull directly from github instead of packagist. I needed them (or did I?) when I was using the feature branch. Could be removed now, I think, but I was planning to wait until changing to a release number.

forevermatt: I think you can simply specify something like `dev-feature/my-feature-name` as the required version, without having to do the `repositories` approach (for what it's worth).

https://getcomposer.org/doc/articles/versions.md#branches

PR #107, opened 2020-04-21
Update SimpleSAMLphp to 1.18.6

Schparky:
Primary purpose is to change logging to stderr.

Also:
- changed from php memcache module to memecache**d** (Note that these serialize differently. It is unknown at this point whether there will be glitches during the switchover.)
- redirect Apache logs to stdout/err
- add logging env vars: `ENABLE_DEBUG`, `LOGGING_LEVEL`, `LOGGING_HANDLER`

Note:
- allow use of the `develop` branch of the silinternational modules until those are updated in the near future

References:
- [SimpleSAMLphp upgrade notes](https://simplesamlphp.org/docs/stable/simplesamlphp-upgrade-notes-1.18)
- [PHP memcached docs](https://www.php.net/manual/en/book.memcached.php). Note the first user-supplied comment.


dockerbuild/ssp-overrides/config.php
```diff
@@ -93,122 +100,293 @@
...
+     * an external URL other than your own domain. This is the default behaviour.
+     *
+     * Set to NULL to disable checking of URLs. DO NOT DO THIS UNLESS YOU KNOW
+     * WHAT YOU ARE DOING!
+     *
+     * Example:
+     *   'trusted.url.domains' => ['sp.example.com', 'app.example.com'],
      */
-    'secretsalt' => $SECRET_SALT,
+    'trusted.url.domains' => [],
```
forevermatt, line 272: Do we know what to expect from changing this from `null` to an empty array?

Schparky: Not exactly. Talked with Steve about it and we'll see how it works on staging. `simplesamlphp-module-sildisco` tests pass with it set this way.

installed-packages.json
```diff
@@ -7,7 +7,7 @@
         },
```
forevermatt, line 7: Out of scope for this PR, but...

What would you think about adjusting the generation of this "installed-packages.json" file to only list the dependencies that we explicitly include in our composer.json? We initially went with including the versions of all installed packages, but I think the signal-to-noise ratio is too poor to be worth it... I wonder if just seeing the version changes of what we explicitly define as our direct dependencies would be better.

Schparky: Sounds good to me.

PR #127, opened 2021-06-15
Update to PHP 7.4

forevermatt:
### Security
- Update to PHP 7.4

### Fixed
- Clean up leftover apt data after install
- Update dependencies

---

I don't know if this will cause issues with any of our simplesaml-module-* repos, but I plan to update them next.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,12 @@ MAINTAINER Phillip Shipley <phillip.shipley@gmail.com>
 
 ENV REFRESHED_AT 2020-06-14
 
-RUN apt-get update -y && \
-    apt-get install -y php-memcached php-gmp && \
-    apt-get clean
+RUN apt-get update -y \
+    && apt-get install -y \
+        php-gmp \
+        php-memcached \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*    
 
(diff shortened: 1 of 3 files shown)


fillup (approved):

Dockerfile
```diff
@@ -1,12 +1,15 @@
-FROM silintl/php7:7.2
+FROM silintl/php7:7.4
 
 MAINTAINER Phillip Shipley <phillip.shipley@gmail.com>
```
Schparky, line 3:
```suggestion
LABEL maintainer "Phillip Shipley <phillip.shipley@gmail.com>"
```

Schparky (approved):

Dockerfile
```diff
@@ -1,12 +1,15 @@
...
 
-ENV REFRESHED_AT 2020-04-16
+ENV REFRESHED_AT 2020-06-14
```
Baggerone, line 5: 2021?  :-) 
(I wonder if we missed that on other PR's, but I guess it may not matter too much.)

Baggerone (approved):

Dockerfile
forevermatt: Good catch 🙂

PR #137, opened 2021-07-27
Stop tracking builds as deployments

forevermatt:
### Fixed
- Remove CI/CD step to track deployments, since this doesn't directly deploy anywhere


forevermatt:
> Sorry I misled you

No worries. It was a good reminder to have the deployment tracker on my radar, in case it is applicable.

PR #147, opened 2022-01-14
Merge translation overrides

forevermatt:
### Added
- Add PHP file for applying dictionary overrides (in child Docker images)
- Add tests for dictionary overrides
- Document how to use this new merge-dictionaries feature

### Fixed
- Update dependencies
- Only show test browser page details after _failed_ tests (not pending, etc.)
- Fix header levels in README


Baggerone (approved):

Schparky (approved):
nice work

jason-jackson:
I'm curious if there's a way we could do this automatically, without needing to run a script.

forevermatt:
> I'm curious if there's a way we could do this automatically, without needing to run a script.

Not that I know of, but there might be.

PR #151, opened 2022-04-12
7.1.1 Composer Update And Local Dev Tweaks

Baggerone:
- Update composer to avoid a downstream composer problem
- Some local dev tweaks to allow for various types of containers to use it as a hub


Makefile
```diff
@@ -4,7 +4,7 @@ ssp: clean
 	docker-compose up -d ssp
 
 hub: clean
-	docker-compose up -d hub sp1 sp2 idp1 idp2
+	docker-compose up -d ssp-hub.local ssp-hub-sp.local sp2 ssp-hub-idp.local idp2
```
forevermatt, line 7: I had thought we would still need `sp1` and `idp1` to be in that list. Do we not need to access those directly from other containers when doing a `make hub`?

forevermatt (approved):
Approved, with a non-blocking question

Makefile
Baggerone: In the docker-compose.yml file, you changed their names last July to ssp-hub-sp.local and ssp-hub-idp.local.  :-)

PR #153, opened 2022-09-30
Get latest releases for our ssp modules and do composer update

Baggerone:


Schparky (approved):
The home stretch

mtompset (approved):
Assuming the numbered ones are also PHP8 versions. :)

jason-jackson (approved):

PR #156, opened 2022-11-21
Add temporary Google Analytics 4 ID config

Baggerone:


dockerbuild/ssp-overrides/config.php
```diff
@@ -70,6 +70,7 @@
 $GOOGLE_ENABLE = Env::get('GOOGLE_ENABLE', false);
 $HUB_MODE = Env::get('HUB_MODE', false);
 $ANALYTICS_ID = Env::get('ANALYTICS_ID', null);
+$ANALYTICS_ID_GA4 = Env::get('$ANALYTICS_ID_GA4', null);
```
forevermatt, line 73: I think the default value for `Env::get()` is already `null`, so you can remove the 2nd parameter if you want to. I don't know why it was included on the call above as well.

forevermatt (approved):

Schparky (approved):

Baggerone:
We're just going to do a hard change over and not worry about having both Google Analytics properties gathering data together for a while.

PR #158, opened 2022-11-28
get develop branch of material for testing appsdev idp

Baggerone:


forevermatt (commented):
What would you think of, instead, changing the AppsDev IDP to use this feature branch (in order to test these changes)? I always shy from allowing code that is specifically non-production code (like using a dev branch of one of our dependencies) into the `develop` branch of our code, which is in theory meant to be code that is ready to deploy at any time.

Baggerone:
> What would you think of, instead, changing the AppsDev IDP to use this feature branch (in order to test these changes)? I always shy from allowing code that is specifically non-production code (like using a dev branch of one of our dependencies) into the `develop` branch of our code, which is in theory meant to be code that is ready to deploy at any time.

PR #175, opened 2023-05-16
Use local database for testing sql session store type

Baggerone:
I over wrote the idp1 config file with your newly modified config file and added the necessary env vars to make it use the sql session store type.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,6 +17,16 @@ services:
     env_file:
       - ./local.env
 
+  db:
+    image: mariadb:10
+    ports:
+      - "3306"
+    environment:
+      MYSQL_ROOT_PASSWORD: r00tp@ss!
+      MYSQL_DATABASE: silauth
+      MYSQL_USER: silauth
+      MYSQL_PASSWORD: silauth
+
   test:
     build: .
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -124,6 +136,11 @@ services:
       SECURE_COOKIE: "false"
       SHOW_SAML_ERRORS: "true"
       THEME_USE: "material:material"
+      SESSION_STORE_TYPE: "sql"
+      MYSQL_HOST: "db"
+      MYSQL_DATABASE: "silauth"
+      MYSQL_USER: "silauth"
+      MYSQL_PASSWORD: "silauth"
```
briskt, line 143: Should this match the password on line 25?

briskt (commented):
Should the passwords match in docker-compose.yml?

docker-compose.yml
Baggerone: Line 25 is `MYSQL_ROOT_PASSWORD`. I assume it's for the root user rather than for the `silauth` user.

I took this from simplesamlphp-module-silauth/docker-compose.yml.

briskt (approved):

PR #177, opened 2023-06-06
Get latest material version for showing name of IdP on TOTP page

Baggerone:
[IDP-31](https://itse.youtrack.cloud/issue/IDP-31) In IDP 2FA prompt, remind the user which account they're getting the code for


mtompset (approved):
What could version updates do? Break logins for an hour? -- Sorry, bad recent experience with SAML2 dependencies in MailAdmin. :)

PR #183, opened 2023-07-31
don't check id-broker status in this service's status check

briskt:
### Fixed
- Updated dependencies, especially [silinternational/simplesamlphp-module-silauth 7.1.1](https://github.com/silinternational/simplesamlphp-module-silauth/releases/tag/7.1.1)
- Use a simpler dependency list format

### DIFF
--- a/installed-packages.json
+++ b/installed-packages.json
@@ -1,148 +1,448 @@
 {
-    "installed": [
-        {
-            "name": "behat/behat",
-            "direct-dependency": true,
-            "homepage": "http://behat.org/",
-            "source": "https://github.com/Behat/Behat/tree/v3.13.0",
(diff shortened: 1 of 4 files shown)


mtompset (approved):
Hope the installed-packages.json format change doesn't affect other code elsewhere.

Baggerone (approved):

installed-packages.json
forevermatt, line 1: Hmm... I'm realizing my sample code for this didn't actually result in valid JSON... though it's close. I wonder if that matters. 🤷

forevermatt (approved):

installed-packages.json
briskt: Not really a problem, but it turned out to be an easy fix.

PR #188, opened 2024-03-08
IDP-892 - enable modules using module.enable in config.php

briskt:
### Added
- Added all required modules to the `module.enable` section of the config.php file. ([IDP-892](https://itse.youtrack.cloud/issue/IDP-892))

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,11 +36,7 @@ WORKDIR /data
 # Install/cleanup composer dependencies
 COPY composer.json /data/
 COPY composer.lock /data/
-# TODO/FIXME: Disabled the self-update due to a breaking change between composer 2.6.6 and 2.7.1 that affects the
-# loading of the simplesamlphp/simplesamlphp/modules folder. The Docker build fails on the sildisco/sspoverrides line.
-# It is not well understood what changed in composer, but since the overrides will need to be redesigned during
-# the SimpleSAMLphp 2.x upgrade, this issue is deferred until then.
-#RUN composer self-update --no-interaction
+RUN composer self-update --no-interaction
 RUN COMPOSER_ALLOW_SUPERUSER=1 composer install --prefer-dist --no-interaction --no-dev --optimize-autoloader --no-scripts --no-progress
 
 # Copy in SSP override files
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN apt-get update -y \
         php-gmp \
         php-memcached \
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*    
+    && rm -rf /var/lib/apt/lists/*
 
 # Create required directories
 RUN mkdir -p /data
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -36,12 +36,8 @@ WORKDIR /data
...
 COPY composer.json /data/
 COPY composer.lock /data/
-# TODO/FIXME: Disabled the self-update due to a breaking change between composer 2.6.6 and 2.7.1 that affects the
-# loading of the simplesamlphp/simplesamlphp/modules folder. The Docker build fails on the sildisco/sspoverrides line.
-# It is not well understood what changed in composer, but since the overrides will need to be redesigned during
-# the SimpleSAMLphp 2.x upgrade, this issue is deferred until then.
-#RUN composer self-update --no-interaction
-RUN composer install --prefer-dist --no-interaction --no-dev --optimize-autoloader --no-scripts --no-progress
+RUN composer self-update --no-interaction
+RUN COMPOSER_ALLOW_SUPERUSER=1 composer install --prefer-dist --no-interaction --no-dev --optimize-autoloader --no-scripts --no-progress
```
jason-jackson, line 40: While this works, I'd prefer if we investigated using another user, as this is not recommended. Possibly `USER www-data`.

jason-jackson, line 40: While this works, I'd prefer if we investigated using another user, as this is not recommended. Possibly `USER www-data`.

jason-jackson (changes requested):

Dockerfile
briskt: Not recommended by whom?

jason-jackson: Basing it off of https://getcomposer.org/doc/faqs/how-to-install-untrusted-packages-safely.md

> Certain Composer commands, including exec, install, and update allow third party code to execute on your system. This is from its "plugins" and "scripts" features. Plugins and scripts have full access to the user account which runs Composer. For this reason, it is strongly advised to avoid running Composer as super-user/root.

But then they also say
> [COMPOSER_ALLOW_SUPERUSER](https://getcomposer.org/doc/03-cli.md#composer-allow-superuser)
>
> If set to 1, this env disables the warning about running commands as root/super user. It also disables automatic clearing of sudo sessions, so you should really only set this if you use Composer as a super user at all times like in docker containers.

briskt: I think all of the warnings are meant for situations other than Docker. Especially during the Docker build, which is a very controlled environment.

briskt: @forevermatt I think @jason-jackson would like your opinion on this

forevermatt: I think it would be good to be running as a non-root user, as Jason suggested, with something like the `USER www-data` command, though I haven't researched all of the specifics (at least not recently).

forevermatt: Are there other places where we are using `COMPOSER_ALLOW_SUPERUSER=1` in production? (I suspect you're right, that we are... at the time of adding that to a codebase seems like an appropriate time to reevaluate whether we need to be bypassing that security recommendation, though.)

However, I do not have a very solid understanding of the risk of running as `root` in a docker container, just a general recollection (right or wrong) that it's not recommended, and that a simple `USER www-data` would fix it. If that's not the case, we can certainly proceed with this `COMPOSER_ALLOW_SUPERUSER` in place.

@briskt Thoughts?

briskt: I'm not aware of a single place we use anything other than the root user to build a Docker image. In fact, I'm not sure it's even possible to build an image using a non-root user.

briskt: I just checked the meaning of the `USER` instruction because I wasn't clear on whether it changed the build user or the runtime user. It does both, so it is possible, but I still am not aware of any place we have done this except in development.

forevermatt: Okay, thank you for looking into it. I'm fine with proceeding as-is. We may want to add an issue to our backlog for investigating this further at some point.

forevermatt: https://itse.youtrack.cloud/issue/ITSE-1158

PR #191, opened 2024-03-28
remove ADMIN_PROTECT_INDEX_PAGE option

briskt:
### Removed
- Remove the `ADMIN_PROTECT_INDEX_PAGE` option. The `Logout page` feature had to be adjusted, but that was the only place that expected the value to be `false`.


briskt:
Tests in other projects depend on this option. Closing this PR until those tests can be updated.

PR #195, opened 2024-05-10
directly include the mfa module 

briskt:
### Added
- Copied all pertinent code, configuration, and tests from [silinternational/simplesamlphp-module-mfa](https://github.com/silinternational/simplesamlphp-module-mfa).
- Added a new container `pwmanager` to help with testing.

### Removed
- Removed empty placeholder test scenarios.

### Changed
- Changed the namespace on the `Assert`, `LoggerFactory` and `LoginBrowser` classes to avoid a custom autoload.

--

_Note: the [first](https://github.com/silinternational/ssp-base/commit/09380a65677ac27eb5acf32799852a23247b1877) commit is mostly just copied files, and the remainder are the changes made to those files._


composer.json
```diff
@@ -35,7 +36,10 @@
   "autoload": {
     "files": [
       "vendor/yiisoft/yii2/Yii.php"
-    ]
+    ],
+    "psr-4": {
+      "Sil\\SspMfa\\Behat\\": "features/"
```
forevermatt, line 41: Is this namespace still correct, given that those files are now here in ssp-base? Looks like now it might need to be `"SimpleSAML\\Module\\mfa\\Behat\\"`, but I'm not positive.

development/idp-local/config/authsources.php
```diff
@@ -77,7 +91,7 @@
                 gmdate('YmdHis\Z', strtotime('+6 months')),
             ],
             'mfa' => [
-                'prompt' => 'yes',
+                'prompt' => 'no',
```
forevermatt, line 94: Why this change (and the other `mfa.prompt` changes from "yes" to "no")? (I presume they are legitimately needed, I'd simply like to learn why, to better understand.)

features/bootstrap/MfaContext.php
```diff
@@ -1,58 +1,23 @@
 <?php
-namespace Sil\SspMfa\Behat\context;
```
forevermatt: It looks like we completely lost the namespace here. Was that intentional?

features/bootstrap/MfaContext.php
```diff
@@ -582,7 +541,7 @@ public function theUsersBrowserSupportsUf()
             'Update USER_AGENT_WITH_WEBAUTHN_SUPPORT to a User Agent with WebAuthn support'
         );
         
-        $this->driver->getClient()->setServerParameter('HTTP_USER_AGENT', $userAgentWithWebAuthn);
+//        $this->driver->getClient()->setServerParameter('HTTP_USER_AGENT', $userAgentWithWebAuthn);
```
forevermatt, line 544: This seems like a significant change... I think that (now commented-out) line was what set the user agent to a string expected to indicate WebAuthn support. Was this change intended to be temporary?

features/bootstrap/MfaContext.php
```diff
@@ -716,7 +675,7 @@ public function theUsersBrowserDoesNotSupportUf()
             'Update USER_AGENT_WITHOUT_WEBAUTHN_SUPPORT to a User Agent without WebAuthn support'
         );
         
-        $this->driver->getClient()->setServerParameter('HTTP_USER_AGENT', $userAgentWithoutWebAuthn);
+//        $this->driver->getClient()->setServerParameter('HTTP_USER_AGENT', $userAgentWithoutWebAuthn);
```
forevermatt, line 678: This also looks like it would break some tests. Why comment it out? (Thanks for helping me understand.)

features/mfa.feature
```diff
@@ -137,13 +141,13 @@ Feature: Prompt for MFA credentials
...
-      |           |   TOTP   |                | does not support WebAuthn |   TOTP           |
-      |           |   TOTP   | , backup codes | does not support WebAuthn |   TOTP           |
-      |           |          |   backup codes | does not support WebAuthn |      backup code |
+#      | WebAuthn  |          |                | does not support WebAuthn | WebAuthn         |
+#      | WebAuthn  | , TOTP   |                | does not support WebAuthn |   TOTP           |
+#      | WebAuthn  |          | , backup codes | does not support WebAuthn |      backup code |
+#      | WebAuthn  | , TOTP   | , backup codes | does not support WebAuthn |   TOTP           |
+#      |           |   TOTP   |                | does not support WebAuthn |   TOTP           |
+#      |           |   TOTP   | , backup codes | does not support WebAuthn |   TOTP           |
+#      |           |          |   backup codes | does not support WebAuthn |      backup code |
```
forevermatt, line 150: This seems related to my previous two comments. Is lack-of-WebAuthn-support something that's (essentially) not a thing any more?

features/mfa.feature
```diff
@@ -212,6 +216,7 @@ Feature: Prompt for MFA credentials
     Given I provide credentials that have a manager code
       And I login
     When I submit the correct manager code
+      And I click the remind-me-later button
```
forevermatt, line 219: This seems unrelated to what this scenario is testing. I wonder if it's an artifact of an accidentally-near password-expiry date. Was this copied from the mfa module, or was it needed to fix the tests after adding in the mfa module's code and tests?

(Similar question with the other place this line was added.)

forevermatt:
Thanks for doing this, @briskt. It looks like it was a fair bit of work to merge this in. 👍

development/idp-local/config/authsources.php
briskt: As far as I can tell, the value of `prompt` was a "don't care" before the modules were merged together. In the simplesamlphp-module-profilereview repo (where these test users came from), the mfa module wasn't enabled, so the `mfa.prompt` attribute was ignored. Now that all modules are enabled during test, this needs to be set to `no` so as to not interfere with the `profilereview` feature tests.

features/bootstrap/MfaContext.php
briskt: I did this according to my understanding of the Behat documentation.

features/bootstrap/MfaContext.php
briskt: Yes, it is temporary. I spent many long and frustrating hours trying to find a browser emulator driver that would support this. As you have seen in the feature file, all of the tests that simulate a browser that does not support webauthn have been disabled.

The longer answer is that the original Mink driver that was used for these tests, Goutte, has been archived in favor of Browser Kit. I was unable to get that to work -- the only documentation I could find was for other frameworks. The Chrome driver doesn't seem to have a way to simulate a different user agent.

features/mfa.feature
briskt: It was a bit of (re-)education for me. Apparently we prompt for a profile review when a user requires the use of a Manager MFA code. Since both modules are enabled in these tests, it required this additional step.

composer.json
briskt: Good question. Actually, I think it can be anything as long as the usage matches. I'll change it to better match where it now lives.

features/bootstrap/MfaContext.php
briskt: Looking back at the [docs](https://behat.org/en/latest/user_guide/context.html#context-class-requirements) it doesn't explicitly say to omit the namespace, but it does mention an autoloader defined for the bootstrap directory.

briskt:
> Thanks for doing this, @briskt. It looks like it was a fair bit of work to merge this in. 👍

It certainly was challenging, especially the part about changing the user agent. And dealing with PHP library versions. Just makes me appreciate Go all the more. :)

features/mfa.feature
forevermatt: Ahh, good to know. I don't remember being aware of that either. Thanks.

composer.json
```diff
@@ -38,7 +38,7 @@
       "vendor/yiisoft/yii2/Yii.php"
     ],
     "psr-4": {
-      "Sil\\SspMfa\\Behat\\": "features/"
+      "SilInternational\\SspBase\\Features\\": "features/"
```
forevermatt, line 41: If you want to keep it shorter, I think we've sometimes/often been using `Sil\` as the root of our class namespaces, rather than `SilInternational\`. Just a option. As long as what's here matches the `namespace` line in the files under "features/", I think it will work just fine (but you probably know better than I do whether that's correct).

features/bootstrap/MfaContext.php
forevermatt: Okay, thanks. I'm sorry it took so many frustrating hours to sort through that.

Since it looks like lack-of-WebAuthn-support is becoming a thing of the past, we could eventually probably remove a lot of this (tests and code) that is focused on detecting WebAuthn support. (Later, not needed right now, and probably after some more discussion.)

Thanks again for your work on this.

forevermatt (approved):

forevermatt:
If you want to keep the class paths shorter, I think we've sometimes/often been using `Sil\` as the root of our class namespaces, rather than `SilInternational\`. Just a option. As long as what's here matches the namespace line in the files under "features/", I think it will work just fine (but you probably know better than I do whether that's correct).

briskt:
> If you want to keep the class paths shorter, I think we've sometimes/often been using `Sil\` as the root of our class namespaces, rather than `SilInternational\`. Just a option. As long as what's here matches the namespace line in the files under "features/", I think it will work just fine (but you probably know better than I do whether that's correct).

I was going for a direct match between the repo name and the namespace, but consistency is often better than "correctness". It's likely a bad assumption that a namespace should match the repository name. I'll change it to "Sil".

PR #196, opened 2024-05-14
directly include the silauth module

briskt:
### Added
- Copied all pertinent code, configuration, and tests from [silinternational/simplesamlphp-module-silauth](https://github.com/silinternational/simplesamlphp-module-silauth).
- Added new containers `broker`, `brokerDb`, and `db` to help with testing.

### Changed
- Changed the namespace on the silauth module source to adhere to SimpleSAMLphp recommendations.

--

_Note: the [first](https://github.com/silinternational/ssp-base/commit/14bb76670f9a60451b65fc363c92437e061f3ac0) commit is mostly just copied files, and the remainder are the changes made to those files._


forevermatt (approved):
Approved.

As a side note, it looks like the action-services.yml and docker-compose.yml are getting pretty long. If they're essentially identical, I'm fine with us working towards combining them. Just my 2 cents (since I was previously the one resistant to that idea).

jason-jackson:
> Approved.
> 
> As a side note, it looks like the action-services.yml and docker-compose.yml are getting pretty long. If they're essentially identical, I'm fine with us working towards combining them. Just my 2 cents (since I was previously the one resistant to that idea).

I think if we combine them we need to still handle getting the environment variables. We could do this by setting the test variables in the local.env.dist file or making a test.env file and renaming whichever file we use in the GH action

briskt:
> Approved.
> 
> As a side note, it looks like the action-services.yml and docker-compose.yml are getting pretty long. If they're essentially identical, I'm fine with us working towards combining them. Just my 2 cents (since I was previously the one resistant to that idea).

Thanks for that comment. I think it may be warranted as well, once higher-priority work has been done.

PR #198, opened 2024-05-22
directly include the material module

briskt:
### Added
- Copied all pertinent code and documentation from [silinternational/simplesamlphp-module-material](https://github.com/silinternational/simplesamlphp-module-material).

_Note: [this commit](https://github.com/silinternational/ssp-base/pull/198/commits/42160b1ca17b7fd58afc05e763e2bfe4379eea74) and [this commit](https://github.com/silinternational/ssp-base/pull/198/commits/488245694b0740e5203ce9145f23065daa3a4b1b) are directly copied from the other repo, and the remainder are the changes made to those files._


hobbitronics (commented):
Good to see all the documentation. When I try to run after setting up /etc/hosts and local.env I get `WARN[0000] The "COMPOSER_CACHE_DIR" variable is not set. Defaulting to a blank string. ` even though its set to `/composer`.

briskt:
> Good to see all the documentation. When I try to run after setting up /etc/hosts and local.env I get `WARN[0000] The "COMPOSER_CACHE_DIR" variable is not set. Defaulting to a blank string. ` even though its set to `/composer`.

You may need PHP and Composer on your machine, but I think you can fake it by creating a cache directory and setting `COMPOSER_CACHE_DIR` to the full path to that directory. It must be set on your machine. Setting it in local.env is not enough, I think.

hobbitronics (approved):
Approved based on our chat about fixing the UserPass.php and default-enable files getting created and documenting the COMPOSER_CACHE_DIR variable not being picked up without it being set in your local environment or before the make command.
image

hobbitronics:
> -[] Approved based on our chat about fixing the UserPass.php and default-enable files getting created and documenting the COMPOSER_CACHE_DIR variable not being picked up without it being set in your local environment or before the make command. image

-[] ..and adding cp local.broker.env.dist ...

PR #206, opened 2024-05-27
consistent use of maps for docker-compose environment

briskt:
### Fixed
- Make consistent the use of YAML map for environment variables in docker compose

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -345,10 +345,10 @@ services:
       ADMIN_EMAIL: john_doe@there.com
       ADMIN_PASS: sp3
       SECRET_SALT: h57fjemb&dn^nsJFGNjweJz3
-      SECURE_COOKIE: false
-      SHOW_SAML_ERRORS: true
-      SAML20_IDP_ENABLE: false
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -364,15 +364,15 @@ services:
...
-      - THEME_USE=default
+      ADMIN_EMAIL: john_doe@there.com
+      ADMIN_PASS: sp1
+      IDP_NAME: THIS VARIABLE IS REQUIRED BUT PROBABLY NOT USED
+      SECRET_SALT: NOT-a-secret-k49fjfkw73hjf9t87wjiw
+      SECURE_COOKIE: "false"
+      SHOW_SAML_ERRORS: "true"
+      SAML20_IDP_ENABLE: "false"
+      ADMIN_PROTECT_INDEX_PAGE: "false"
+      THEME_USE: material:material
```
hobbitronics, line 375: Just checking that this change was intentional.

briskt: Not really. It should have been in the next PR. But since tests pass, it's ok to leave it like this.

PR #211, opened 2024-06-04
development and test adjustments

briskt:
### Removed
- Removed Xdebug from the default development environment

### Added
- Added instructions to temporarily set up development environment for Xdebug

### Changed
- Use a Docker argument to control the flags for `composer install`. This provides a way to include dev dependencies in the test container without running `composer install` at the time tests are run.


forevermatt (approved):
I don't fully understand why we don't need the `composer install` in `run-integration-tests.sh`, but overall this seems good.

briskt:
> I don't fully understand why we don't need the `composer install` in `run-integration-tests.sh`, but overall this seems good.

I think it's only to install behat. If there's a better way, I'd love to know about it.

PR #213, opened 2024-06-06
share the use of static functions in Mfa authproc

briskt:


jason-jackson (approved):
It looks fine, but it feels a little odd to have one module import the another (this could just be me, and it's not a big issue at all - I'm totally fine with doing it). A couple questions/alternatives:

1. Should the methods be moved out into a helper class instead?  (Issue with this is `where would it go?`)
2. Should ProfileReview extend module? (Issue being does it need to have all the other junk)

So overall, I think it's fine as is unless you have a good answer to any of the above questions. As I'm more concerned with the issues I have with alternatives than I am with this way.

briskt:
Decided that the duplication is less of a problem than the module interdependency.

PR #216, opened 2024-06-12
Release 10.0.0-alpha.2 -- cleanup, twig prep, typehints

briskt:
### Added
- Added an announcement file for local development
- Added a test feature that logs the html and saves a screenshot on failure

### Changed (breaking)
- Always use the `material` theme
- Added PHP typehints to module code

### Changed (non-breaking)
- Moved PHP code out of templates in preparation for conversion to Twig templating
- Changed the name of the Actions test container to `test`
- Include dev dependencies (e.g. behat) in the test image
- Made the `ADMIN_EMAIL` variable no longer required
- Updated files "lifted" from the SimpleSAMLphp library

### Removed
- Removed `default` theme files
- Removed xdebug from the local development image
- Removed old `prompt-for-mfa-u2f.php` template file
- Removed duplicate copy of simplewebauthn browser.js file

### Fixed
- Fixed a "not found" error on the simplewebauthn browser.js file, which prevented use of a Webauthn 2sv option at login
- Fixed failure in hub mode due to a missing dependency (AWS SDK)
- Fixed failure to ~run~ complete migrations. Reverted the migration namespace back to the original (\Sil\SilAuth\migrations).
- Better error handling for manager "rescue" code request failure


briskt:
Most changes have been reviewed on a feature branch. Recent changes, such as bugs found while testing on the appsdev IdP and staging hub were committed to the develop branch. Reference [this diff](https://github.com/silinternational/ssp-base/compare/5d884b..b33f09) to see those changes isolated.

.github/workflows/test-and-publish.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
       - name: test
-        run: docker-compose -f actions-services.yml run --rm app ./run-tests.sh
+        run: docker-compose -f actions-services.yml run --rm test ./run-tests.sh
```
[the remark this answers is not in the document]
briskt: Thanks for the reminder, but I've been aware of that for several years now. It's somewhere down the list of things to address. By the way, by adding this comment you took the php-devs group off of the review list. I had to go add it back.

development/idp-local/config/config.php
```diff
@@ -22,7 +22,7 @@
 
 try {
     // Required to be defined in environment variables
-    $ADMIN_EMAIL = Env::requireEnv('ADMIN_EMAIL');
+    $ADMIN_EMAIL = Env::get('ADMIN_EMAIL', 'na@example.org');
```
mtompset, line 25: Why this change?

development/idp-local/config/config.php
```diff
@@ -22,7 +22,7 @@
 
 try {
     // Required to be defined in environment variables
```
mtompset, line 24: Comment is no longer accurate

development/idp-local/config/config.php
```diff
@@ -964,7 +963,7 @@
     /*
      * Which theme directory should be used?
      */
-    'theme.use' => $THEME_USE,
+    'theme.use' => 'material:material',
```
mtompset, line 966: Why move from variable to hard-coded?

development/idp2-local/config/config.php
```diff
@@ -22,7 +22,7 @@
 
 try {
     // Required to be defined in environment variables
```
mtompset, line 24: Comment is inaccurate.

development/idp2-local/config/config.php
```diff
@@ -964,7 +963,7 @@
     /*
      * Which theme directory should be used?
      */
-    'theme.use' => $THEME_USE,
+    'theme.use' => 'material:material',
```
mtompset, line 966: Having the variable makes more sense to me. Why this change again?

docker-compose.yml
```diff
@@ -19,7 +19,10 @@ services:
       PMA_PASSWORD: silauth
 
   test:
-    build: .
+    build:
+      context: .
+      args:
```
mtompset, line 24: This was cool to learn about.

development/idp-local/config/config.php
forevermatt, line 25: As @mtompset noticed, since this one isn't required now, it doesn't match the comment explaining this section of the code. Could you move this `ADMIN_EMAIL` line down to somewhere around line 47, to be with the other optional env. vars?

Similarly in the other places an `Env::requireEnv()` was changed to an `Env::get()` with a fallback/default value.

features/bootstrap/FeatureContext.php
```diff
@@ -42,11 +44,27 @@ public function __construct()
     /** @AfterStep */
     public function afterStep(AfterStepScope $scope)
     {
-        if (! $scope->getTestResult()->getResultCode() === StepResult::FAILED) {
+        if ($scope->getTestResult()->getResultCode() === StepResult::FAILED) {
             $this->showPageDetails();
+            $this->takeScreenshot();
```
forevermatt, line 49: Nice! 😁

modules/mfa/www/prompt-for-mfa.php
```diff
@@ -124,7 +125,7 @@
 $t->data['stateId'] = $stateId;
 $t->data['supportsWebAuthn'] = LoginBrowser::supportsWebAuthn($userAgent);
 $browserJsHash = md5_file(__DIR__ . '/simplewebauthn/browser.js');
-$t->data['browserJsPath'] = 'simplewebauthn/browser.js?v=' . $browserJsHash;
+$t->data['browserJsPath'] = '/module.php/mfa/simplewebauthn/browser.js?v=' . $browserJsHash;
```
forevermatt, line 128: Was the path still correct in the line calculating the `$browserJsHash`?

forevermatt (approved):
Approved, with very minor (non-blocking) comments or questions. Thanks! 👍

modules/mfa/www/prompt-for-mfa.php
briskt: Yeah, it's a bit confusing because the path for the controller is different than the path that the browser sees.

development/idp-local/config/config.php
briskt: Because we never actually use the default template, which is the only other option for our custom modules. It only led to confusion because some tests used "default" and others used "material". Plus, it allowed me to delete a bunch of files related to the default template.

development/idp2-local/config/config.php
briskt: Because we never actually use the default template, which is the only other option for our custom modules. It only led to confusion because some tests used "default" and others used "material". Plus, it allowed me to delete a bunch of files related to the default template.

PR #218, opened 2024-06-13
remove GOOGLE_ENABLE

briskt:
### Removed
- Removed `GOOGLE_ENABLE` option since we never use it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -268,14 +268,6 @@ services:
 
       # Enable checking our test metadata
       - ./dockerbuild/run-metadata-tests.sh:/data/run-metadata-tests.sh
-
-      # Local modules
-      - ./modules/mfa:/data/vendor/simplesamlphp/simplesamlphp/modules/mfa
-      - ./modules/expirychecker:/data/vendor/simplesamlphp/simplesamlphp/modules/expirychecker
-      - ./modules/profilereview:/data/vendor/simplesamlphp/simplesamlphp/modules/profilereview
-      - ./modules/silauth:/data/vendor/simplesamlphp/simplesamlphp/modules/silauth
-      - ./modules/sildisco:/data/vendor/simplesamlphp/simplesamlphp/modules/sildisco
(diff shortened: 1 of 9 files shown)


docker-compose.yml
```diff
@@ -298,14 +290,6 @@ services:
...
       # Utilize custom metadata
       - ./development/sp2-local/metadata/saml20-idp-remote.php:/data/vendor/simplesamlphp/simplesamlphp/metadata/saml20-idp-remote.php
-
-      # Local modules
-      - ./modules/mfa:/data/vendor/simplesamlphp/simplesamlphp/modules/mfa
-      - ./modules/expirychecker:/data/vendor/simplesamlphp/simplesamlphp/modules/expirychecker
-      - ./modules/profilereview:/data/vendor/simplesamlphp/simplesamlphp/modules/profilereview
-      - ./modules/silauth:/data/vendor/simplesamlphp/simplesamlphp/modules/silauth
-      - ./modules/sildisco:/data/vendor/simplesamlphp/simplesamlphp/modules/sildisco
-      - ./modules/material:/data/vendor/simplesamlphp/simplesamlphp/modules/material
```
mtompset, line 308: Are you sure about these local modules being removed?

briskt: Yes, I shouldn't have added them to begin with. The SP containers are pretty bare-bones. They are only there to receive login requests and redirect to the Hub and/or IdP containers. Our custom modules are all for IdP and Hub use.

hobbitronics (approved):
I'm not sure what enabling the "core", "saml", and "admin" modules does, but assume they won't break anything.

briskt:
> I'm not sure what enabling the "core", "saml", and "admin" modules does, but assume they won't break anything.

The old way, in SimpleSAMLphp 1.x, was to include a file named `enable` or `default-enable` in the module directory. The new way, in SimpleSAMLphp 2.x, is to use this configuration. These three are the standard (non-custom) modules that we use.

PR #220, opened 2024-06-13
remove TIMEZONE config option

briskt:

### DIFF
--- a/development/idp-local/config/config.php
+++ b/development/idp-local/config/config.php
@@ -46,7 +46,6 @@ $ADMIN_EMAIL = Env::get('ADMIN_EMAIL', 'na@example.org');
 $ADMIN_NAME = Env::get('ADMIN_NAME', 'SAML Admin');
 $ADMIN_PROTECT_INDEX_PAGE = Env::get('ADMIN_PROTECT_INDEX_PAGE', true);
 $SHOW_SAML_ERRORS = Env::get('SHOW_SAML_ERRORS', false);
-$TIMEZONE = Env::get('TIMEZONE', 'GMT');
 $ENABLE_DEBUG = Env::get('ENABLE_DEBUG', false);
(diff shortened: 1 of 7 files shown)


development/idp-local/config/config.php
```diff
@@ -46,7 +46,6 @@
 $ADMIN_NAME = Env::get('ADMIN_NAME', 'SAML Admin');
 $ADMIN_PROTECT_INDEX_PAGE = Env::get('ADMIN_PROTECT_INDEX_PAGE', true);
 $SHOW_SAML_ERRORS = Env::get('SHOW_SAML_ERRORS', false);
-$TIMEZONE = Env::get('TIMEZONE', 'GMT');
```
mtompset, line 49: Wouldn't it have been easier to hardcode this, rather than change all 'timezone' references below?

development/idp2-local/config/config.php
```diff
@@ -46,7 +46,6 @@
 $ADMIN_NAME = Env::get('ADMIN_NAME', 'SAML Admin');
 $ADMIN_PROTECT_INDEX_PAGE = Env::get('ADMIN_PROTECT_INDEX_PAGE', true);
 $SHOW_SAML_ERRORS = Env::get('SHOW_SAML_ERRORS', false);
-$TIMEZONE = Env::get('TIMEZONE', 'GMT');
```
mtompset, line 49: Wouldn't it have been easier to hardcode this, rather than change all 'timezone' references below? What if the timezone code changes from GMT to something else in the future? More code changes.

development/idp-local/config/config.php
briskt: Perhaps

development/idp2-local/config/config.php
briskt: I sure hope it doesn't change. But you're right about it being easier the other way.

briskt:
@mtompset I'm not sure why you didn't approve this. I'm merging anyway.

PR #221, opened 2024-06-13
remove memcache option

briskt:
### Removed
- Removed memcache option for session storage

### DIFF
--- a/development/idp-local/config/config.php
+++ b/development/idp-local/config/config.php
@@ -56,6 +56,7 @@ $THEME_COLOR_SCHEME = Env::get('THEME_COLOR_SCHEME', null);
 
 $SECURE_COOKIE = Env::get('SECURE_COOKIE', true);
 $SESSION_DURATION = (int)(Env::get('SESSION_DURATION', (60 * 60 * 10))); // 10 hours.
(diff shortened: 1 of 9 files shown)


development/idp-local/config/config.php
```diff
@@ -806,20 +802,13 @@
...
-            ],
-        ],
-        [
-            [
-                'hostname' => $MEMCACHE_HOST2,
-                'port' => $MEMCACHE_HOST2_PORT,
-            ],
-        ],
-    ],
+//    'memcache_store.servers' => [
```
mtompset, line 805: Why not /* */ comment around the outdated code?

jason-jackson: I think doing that would make it too similar to comments detailing what config values do.

jason-jackson (approved):
I don't know remember if this is being used or not and didn't check that. Code looks fine.

development/idp-local/config/config.php
briskt: Not to mention that PhpStorm doesn't use /* */ when you press Ctrl-/

briskt:
> I don't know remember if this is being used or not and didn't check that. Code looks fine.

Hasn't been used since the multiregion project. We decided to save money by using the database instead. No need to pay for Elasticache since we don't need the slight performance benefit in provides.

PR #224, opened 2024-06-14
rename SimpleSAMLphp directories

briskt:
### Changed (breaking)
- Rename all `lib` directories to `src`
- Rename all `www` directories to `public`

To keep this pull request relatively small in scope, tests have been skipped because they are not yet passing.

[Reference](https://simplesamlphp.org/docs/stable/simplesamlphp-upgrade-notes-2.0.html#changes-relevant-for-module-developers)

https://itse.youtrack.cloud/issue/IDP-869


jason-jackson:
Should vendor files be changing too?

briskt:
> Should vendor files be changing too?

Vendor files aren't committed to the repo

dockerbuild/run.sh
```diff
@@ -7,7 +7,7 @@ set -x
 set -e
 
 # This is a temporary fix (bug workaround) until ssp 2.0 is in use
-sed -i 's_\(\\SimpleSAML\\Error\\Assertion::installHandler()\)_// \1 _' /data/vendor/simplesamlphp/simplesamlphp/www/_include.php
+sed -i 's_\(\\SimpleSAML\\Error\\Assertion::installHandler()\)_// \1 _' /data/vendor/simplesamlphp/simplesamlphp/public/_include.php
```
jason-jackson, line 10: Sorry, I'll put the comment here to show what I mean. The vendor reference is changing from `www` to `public` I assume that's fine, but wanted to double-check.

briskt: Yes, that's correct.

PR #226, opened 2024-06-14
Release 10.0.0-alpha.3 -- more cleanup and prep for SSP 2

briskt:
### Added
- Added notes about manual testing to README.

### Changed (non-breaking)
- Added saml, core, and admin modules to the `module.enable` config entry to prepare for SimpleSAMLphp 2.x. ([IDP-892](https://itse.youtrack.cloud/issue/IDP-892))

### Removed
- Removed memcache option for session storage. ([IDP-1041](https://itse.youtrack.cloud/issue/IDP-1041))
- Removed `TIMEZONE` config option. ([IDP-1041](https://itse.youtrack.cloud/issue/IDP-1041))
- Removed `GOOGLE_ENABLE` option. ([IDP-1041](https://itse.youtrack.cloud/issue/IDP-1041))

### Fixed
- Removed the hyphen from "docker compose".
- Removed the version line from actions-services.yml.
- Removed "--rm" from docker compose lines in the GitHub Actions workflow.


docker-compose.yml
```diff
@@ -186,7 +186,7 @@ services:
       MYSQL_DATABASE: "silauth"
       MYSQL_USER: "silauth"
       MYSQL_PASSWORD: "silauth"
-      BASE_URL_PATH: "http://ssp-idp1.local/"
+      BASE_URL_PATH: "http://ssp-idp1.local/" # change this to "http://ssp-idp1.local:8085" for manual browser testing
```
jason-jackson, line 189: It might be worth making this a TODO or FIXME comment

jason-jackson (approved):

docker-compose.yml
mtompset: I think this is more documentary than needing to fix.

mtompset (approved):
So scrolling on a phone to review

docker-compose.yml
briskt: @mtompset is right. However, there should be an action item to resolve the need to make this change. I'll file something in YouTrack

Vagrantfile
```diff
@@ -56,16 +56,16 @@ Vagrant.configure(2) do |config|
     sudo groupadd docker
     sudo usermod -aG docker vagrant
     # Install Docker Compose
-    curl -L https://github.com/docker/compose/releases/download/1.17.0/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
-    chmod +x /usr/local/bin/docker-compose
+    curl -L https://github.com/docker/compose/releases/download/1.17.0/docker compose-`uname -s`-`uname -m` > /usr/local/bin/docker compose
+    chmod +x /usr/local/bin/docker compose
```
forevermatt, line 60: This seems like a bug, since it was creating a binary, but that binary now will attempt to have a space in its name.

I would recommend simply deleting the Vagrantfile, personally. If that's too big a change, we could remove this `curl` call that was installing docker-compose, since it's now included as a docker subcommand.

forevermatt (commented):
Thanks for all the work you did on this. 👍

Vagrantfile
briskt: Oh, that's an excellent observation. I will just remove the Vagrantfile and anything related to it.

PR #229, opened 2024-06-17
SimpleSAML Utils now requires instantiation

briskt:
### Changed
- The SimpleSAML\Utils classes now require instantiation to call member functions because the methods are no longer static.

https://itse.youtrack.cloud/issue/IDP-1103


forevermatt (approved):
Beyond the parentheses issue, though, this all looks fine to me.

PR #231, opened 2024-06-18
use ssp-base 9.3.0 (SimpleSAMLphp 1.19) for test SPs

briskt:
Since the `admin.protectindexpage` (`ADMIN_PROTECT_INDEX_PAGE`) is no longer supported in SSP 2, it may be easiest to just leave the test SPs on version 1.19. This is fine because we don't use ssp-base in non-Hub SP implementations.


docker-compose.yml
```diff
@@ -254,7 +254,7 @@ services:
       IDP_NAME: "IdP3"
 
   ssp-sp1.local:
-    build: .
+    image: silintl/ssp-base:9.3.0
```
forevermatt, line 257: Hmm... this concerns me slightly, though it is a nice simple fix for the problem.

I supposed since this is just to be "some website that supports SAML", for testing the actual IDP and Hub behavior (which _are_ using the latest code), it's probably totally fine. It just _seems_ undesirable, at least at first.

Okay, no objection. Just needed to think through that. 🙂

briskt: Right, I even have intent to use something even simpler than ssp-base for this. Either someone else's SimpleSAMLphp docker image, or one I create just as a plain-and-simple SP implementation.

PR #236, opened 2024-06-20
convert prompt-for-mfa templates to Twig

briskt:
### Changed
- Converted `prompt-for-mfa` templates to Twig

https://itse.youtrack.cloud/issue/IDP-898


modules/material/themes/material/default/other_mfas.twig
```diff
@@ -0,0 +1,27 @@
+{% if otherOptions|length > 0 %}
+  <div layout-children="column" child-spacing="center">
+    {# used type=button to avoid form submission on click since this is just used to display the ul #}
```
hobbitronics, line 3: So nothing happens on the click event? Would it be better to not use a button? I realize this is just copied from the old file.

modules/material/themes/material/default/other_mfas.twig
```diff
@@ -0,0 +1,27 @@
...
+    </button>
+    <ul class="mdl-menu mdl-js-menu mdl-menu--top-left" data-mdl-for="others">
+      {% for option in otherOptions %}
+        {# TODO: Non-interactive elements should not be assigned mouse or keyboard event listeners. #}
+        <li class="mdl-menu__item" onclick="location.href = '{{ option.callback|raw }}'">
+          <span class="mdl-list__item-primary-content">
+            <img
+              class="mdl-list__item-icon"
+              src="{{ option.image }}"
+              {% set alt = '{mfa:' ~ option.type ~ '_icon}' %}
```
hobbitronics, line 17: is the translation an icon? That would seem strange 🤔

modules/material/themes/material/mfa/prompt-for-mfa-manager.twig
```diff
@@ -0,0 +1,83 @@
...
+          </div>
+
+          <script>
+            ga('send', 'event', 'error', 'managercode','{{ errorMessage|e('js')|raw }}');
+          </script>
+        {% endif %}
+
+        <div class="mdl-card__actions" layout-children="row">
+          <span flex></span>
+          <button name="submitMfa" class="mdl-button mdl-button--raised mdl-button--primary">
```
hobbitronics, line 63:
```suggestion
          <button name="submitMfa" type="submit" class="mdl-button mdl-button--raised mdl-button--primary">
```

modules/material/themes/material/mfa/prompt-for-mfa-totp.twig
```diff
@@ -0,0 +1,81 @@
...
+          </div>
+
+          <script>
+            ga('send', 'event', 'error', 'totp', '{{ errorMessage|e('js')|raw }}');
+          </script>
+        {% endif %}
+
+        <div class="mdl-card__actions" layout-children="row">
+          <span flex></span>
+          <button name="submitMfa" class="mdl-button mdl-button--raised mdl-button--primary">
```
hobbitronics, line 61:
```suggestion
          <button name="submitMfa" type="submit" class="mdl-button mdl-button--raised mdl-button--primary">
```

modules/material/themes/material/mfa/prompt-for-mfa-webauthn.twig
```diff
@@ -0,0 +1,148 @@
...
+        {% if supportsWebAuthn %}
+          <div class="mdl-card__title">
+            <p class="mdl-card__subtitle-text">
+              {{ '{mfa:webauthn_instructions}'|trans }}
+            </p>
+          </div>
+        {% else %}
+          <div class="mdl-card__title">
+            <p class="mdl-typography--text-center mdl-color-text--red">
+              {{ '{mfa:webauthn_unsupported}'|trans|raw }}
```
hobbitronics, line 100: why the `raw` here?

hobbitronics (approved):
Looks good, just had a question about '{mfa:webauthn_unsupported}'|trans|raw`

modules/material/themes/material/default/other_mfas.twig
briskt: No, the string ending in "_icon" is just the translation key. For instance, "webauthn_icon" translates to "USB key icon" in English. Note that this is the alt for the image element.

modules/material/themes/material/default/other_mfas.twig
briskt: I don't really know. It doesn't make sense to me either.

modules/material/themes/material/mfa/prompt-for-mfa-webauthn.twig
briskt: Because it avoids escaping that would happen if auto-escape is turned on. I only today realized my assumption was incorrect. There are a number of these that could be removed from several PRs, and I think I'd prefer to do it all at once so I don't have to type the explanation over and over. One of the side effects of having a number of PRs queued up.

PR #241, opened 2024-06-24
convert loginuserpass template to Twig

briskt:
### Changed
- Converted loginuserpass template to Twig
- Use silauth rather than example-userpass on idp2 container for testing login


docker-compose.yml
```diff
@@ -389,6 +415,15 @@ services:
       MYSQL_USER: "user"
       MYSQL_PASSWORD: "pass"
 
+  brokerpma:
+    image: phpmyadmin/phpmyadmin
```
jason-jackson, line 419:
```suggestion
    image: phpmyadmin
```

modules/material/themes/material/silauth/loginuserpass.twig
```diff
@@ -0,0 +1,146 @@
...
+
+  {% include 'header.twig' %}
+
+  <script>
+    function preventDefault(event) {
+      event.preventDefault();
+    }
+  </script>
+
+  {% if siteKey is defined %}
```
jason-jackson, line 14: I am not sure if defined would catch an empty, but still defined siteKey.

```suggestion
  {% if siteKey is not empty %}
```

modules/material/themes/material/silauth/loginuserpass.twig
```diff
@@ -0,0 +1,146 @@
...
+            <label for="password" class="mdl-textfield__label">
+              {{ '{login:label_password}'|trans }}
+            </label>
+
+            <input
+              type="password"
+              required
+              name="password"
+              class="mdl-textfield__input mdl-color-text--black"
+              {% if not username is empty %} autofocus {% endif %}
```
jason-jackson, line 85: Don't care much either way. I think I prefer readability (change), but emphasis on `not` is fine too (as is).

```suggestion
              {% if username is not empty %} autofocus {% endif %}
```

modules/material/themes/material/silauth/loginuserpass.twig
```diff
@@ -0,0 +1,146 @@
...
+            </span>
+          </p>
+
+          <script>
+            ga('send', 'event', 'error',{{ errorCode|e(js) }}, 'message',{{ message|e(js) }});
+          </script>
+        {% endif %}
+
+        <div class="mdl-card__actions" layout-children="row">
+          {% if forgotPasswordUrl is defined %}
```
jason-jackson, line 109: Same reasoning as above.

```suggestion
          {% if forgotPasswordUrl is not empty %}
```

modules/material/themes/material/silauth/loginuserpass.twig
```diff
@@ -0,0 +1,146 @@
...
+              required
+              name="password"
+              class="mdl-textfield__input mdl-color-text--black"
+              {% if not username is empty %} autofocus {% endif %}
+              id="password"
+            >
+          </div>
+        </div>
+
+        {% if errorCode ?? '' == 'WRONGUSERPASS' %}
```
jason-jackson, line 91: What happens if you use this instead?
```
        {% if errorCode == 'WRONGUSERPASS' %}
```

modules/material/themes/material/silauth/loginuserpass.twig
```diff
@@ -0,0 +1,146 @@
...
+          {% set errorMessageKey = errorparams[1] ?? '{login:error_wronguserpass}'|trans %}
```
jason-jackson, line 92: You call trans later and shouldn't call it here.

```suggestion
          {% set errorMessageKey = errorparams[1] ?? '{login:error_wronguserpass}' %}
```

modules/material/themes/material/silauth/loginuserpass.twig
```diff
@@ -0,0 +1,146 @@
...
+          {% set errorMessageTokens = errorparams[2] ?? null %}
+          {% set message = errorMessageKey|trans({errorMessageKey, errorMessageTokens}) %}
+          <p class="mdl-card__supporting-text mdl-color-text--red error">
+            <i class="material-icons">error</i>
+
+            <span class="mdl-textfield mdl-typography--caption">
+              {{ message|e }}
```
jason-jackson, line 99: Optional. I think you could simplify it, but if you think it's clearer as is, feel free to leave it.

```suggestion
          <p class="mdl-card__supporting-text mdl-color-text--red error">
            <i class="material-icons">error</i>

            <span class="mdl-textfield mdl-typography--caption">
              {{ errorMessageKey|trans({errorMessageKey, errorparams[2] ?? null}) }}
```

modules/material/themes/material/silauth/loginuserpass.twig
```diff
@@ -0,0 +1,146 @@
...
+
+          <span flex></span>
+
+          <button class="mdl-button mdl-button--colored mdl-button--raised">
+            {{ '{login:button_login}'|trans }}
+          </button>
+        </div>
+
+        <section layout-children="row" child-spacing="space-around">
+          {% if helpCenterUrl is defined %}
```
jason-jackson, line 127:
```suggestion
          {% if helpCenterUrl is not empty %}
```

modules/material/themes/material/silauth/loginuserpass.twig
```diff
@@ -0,0 +1,146 @@
...
+            <a href="{{ helpCenterUrl }}" target="_blank" class="mdl-button mdl-button--colored mdl-typography--body-2">
+              {{ '{login:help}'|trans }}
+              <i class="material-icons">launch</i>
+            </a>
+          {% endif %}
+
+          {% if profileUrl is defined %}
```
jason-jackson, line 134:
```suggestion
          {% if profileUrl is not empty %}
```

jason-jackson (commented):
I believe `defined` is more like `isset` and not `!empty`. If you have documentation/proof otherwise feel free to keep it.

modules/material/themes/material/silauth/loginuserpass.twig
jason-jackson: Tested this with fiddle, I was correct. Should be changed.

briskt: It has to check 'defined' or have a null coalesce since `siteKey` will not be defined if the out-of-the-box SSP controller is used. Eventually it won't be, but I'm not there yet.

modules/material/themes/material/silauth/loginuserpass.twig
briskt: Same explanation as above :)

modules/material/themes/material/silauth/loginuserpass.twig
briskt: It would fail if errorCode is undefined. Tried that before getting to this point, I think.

modules/material/themes/material/silauth/loginuserpass.twig
briskt: Ah, good find.

modules/material/themes/material/silauth/loginuserpass.twig
briskt: Thanks, but I think it's clear as is.

modules/material/themes/material/silauth/loginuserpass.twig
briskt: I'll add the `not empty` but the `defined` is necessary.

modules/material/themes/material/silauth/loginuserpass.twig
briskt: ditto

modules/material/themes/material/silauth/loginuserpass.twig
jason-jackson: This is fine, noting an alternative in case we want to explore it later: `errorCode|default == 'WRONGUSERPASS'`. This will default it to an empty string if it is not set.

The other alternative is if we don't want it to fail if it is not given is to turn off `strict_variables` for twig (see [here](https://twig.symfony.com/doc/2.x/api.html#environment-options))

briskt: Good to know about that `|default` filter.

I don't know if (or how) we can modify the Twig settings since it's instantiated in SSP code. Not sure I'd want to change `strict_variables` anyway.

jason-jackson: I just checked empty again with regards to defined, and it seems to handle it if it isn't set (even in strict mode) https://fiddle.nette.org/twig/#63de026fe7

briskt: I don't know what's different between SSP and Fiddle, but it doesn't work.

> Caused by: Twig\Error\RuntimeError: Variable "analyticsTrackingId" does not exist.

PR #242, opened 2024-06-24
combine actions-services and docker-compose config

briskt:
### Changed
- Merged actions-services.yml into docker-compose.yml

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,15 +38,11 @@ services:
       MYSQL_DATABASE: silauth
       MYSQL_USER: silauth
       MYSQL_PASSWORD: silauth
-      COMPOSER_CACHE_DIR: /composer
       PROFILE_URL_FOR_TESTS: http://pwmanager.local/module.php/core/authenticate.php?as=ssp-hub
       ADMIN_PASS: b
       SECRET_SALT: abc123
       IDP_NAME: x
     volumes:
-      - ./composer.json:/data/composer.json
-      - ./composer.lock:/data/composer.lock
-      - ${COMPOSER_CACHE_DIR}:/composer
       - ./dockerbuild/run-integration-tests.sh:/data/run-integration-tests.sh
       - ./dockerbuild/run-metadata-tests.sh:/data/run-metadata-tests.sh
       - ./dockerbuild/run-tests.sh:/data/run-tests.sh
@@ -77,7 +73,7 @@ services:
       - ./installed-packages.json:/data/installed-packages.json
       - ./development/update-composer-deps.sh:/data/update-composer-deps.sh
       - ./vendor:/data/vendor
-      - ${COMPOSER_CACHE_DIR}:/composer
+      - composer-cache:/composer
     env_file:
       - ./local.env
     environment:
@@ -85,13 +81,6 @@ services:
 
   ssp-hub.local:
     build: .
-    depends_on:
-      - ssp-idp1.local
-      - ssp-idp2.local
-      - ssp-idp3.local
-      - ssp-sp1.local
-      - ssp-sp2.local
-      - ssp-sp3.local
     volumes:
       # Utilize custom certs
       - ./development/hub/cert:/data/vendor/simplesamlphp/simplesamlphp/cert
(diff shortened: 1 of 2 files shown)


hobbitronics (approved):
Looks good, but how do you get rid of the port number.

briskt:
> Looks good, but how do you get rid of the port number.

I may not understand your question. The last commit on this PR does away with the port number for that one container. Are you asking something else?

jason-jackson (commented):
Looks good. Now to use extension fields to reuse the environment variables.

jason-jackson (approved):

hobbitronics:
> > Looks good, but how do you get rid of the port number.
> 
> I may not understand your question. The last commit on this PR does away with the port number for that one container. Are you asking something else?

Yeah, I was just wondering why you were able to do that.

briskt:
> > > Looks good, but how do you get rid of the port number.
> > 
> > 
> > I may not understand your question. The last commit on this PR does away with the port number for that one container. Are you asking something else?
> 
> Yeah, I was just wondering why you were able to do that.

Ah, yes, good question. I'm not 100% sure, but it may have to do with some of the tests having been disabled. If a problem surfaces at the time I reenable those tests, I'm sure there's a good solution.

development/sp-local/config/authsources-pwmanager.php
```diff
@@ -12,7 +12,7 @@
 
     'mfa-idp' => [
         'saml:SP',
-        'entityID' => 'http://pwmanager.local:8084',
+        'entityID' => 'http://pwmanager.local',
```
forevermatt, line 15: Yeah, given the name of this one (mfa-idp) and the next (mfa-idp-no-port), this one was probably only here for manual local testing/interaction during development, which requires a port number. It might be better to give this some other/different port number for now, but that's certainly not a blocking concern. (My 2 cents.)

docker-compose.yml
```diff
@@ -38,15 +38,11 @@ services:
...
       MYSQL_USER: silauth
       MYSQL_PASSWORD: silauth
-      COMPOSER_CACHE_DIR: /composer
       PROFILE_URL_FOR_TESTS: http://pwmanager.local/module.php/core/authenticate.php?as=ssp-hub
       ADMIN_PASS: b
       SECRET_SALT: abc123
       IDP_NAME: x
     volumes:
-      - ./composer.json:/data/composer.json
-      - ./composer.lock:/data/composer.lock
```
forevermatt, line 48: Just as a heads-up on this change, removing these means that any changes made to those files will require rebuilding this docker image in order to see the effects of those changes when running tests during local development. Maybe not a big deal, but a possible "gotcha" at some point down the road. (I assume this was done to solve some problem/issue with the GitHub Action build, which is fine.) No objection, just an FYI.

docker-compose.yml
```diff
@@ -77,21 +73,14 @@ services:
       - ./installed-packages.json:/data/installed-packages.json
       - ./development/update-composer-deps.sh:/data/update-composer-deps.sh
       - ./vendor:/data/vendor
-      - ${COMPOSER_CACHE_DIR}:/composer
+      - composer-cache:/composer
```
forevermatt, line 76: Nice fix 👍

docker-compose.yml
```diff
@@ -168,16 +157,16 @@ services:
...
-      SECRET_SALT: "h57fjemb&dn^nsJFGNjweJ"
+      SECRET_SALT: "not-secret-h57fjemb&dn^nsJFGNjweJ"
       IDP_NAME: "IDP 1"
       IDP_DOMAIN_NAME: "mfaidp"
       ID_BROKER_ACCESS_TOKEN: "dummy"
       ID_BROKER_ASSERT_VALID_IP: "false"
       ID_BROKER_BASE_URI: "dummy"
       ID_BROKER_TRUSTED_IP_RANGES: "192.168.0.1/8"
-      MFA_SETUP_URL: "http://pwmanager.local:8084/module.php/core/authenticate.php?as=ssp-hub-custom-port"
+      MFA_SETUP_URL: "http://pwmanager.local/module.php/core/authenticate.php?as=ssp-hub-custom-port"
```
forevermatt, line 167: We'll probably need to find a way to re-add (or resolve the need for) these custom ports later, just as an FYI, so that it can correctly identify whether the user is headed to set up their MFA (to avoid infinitely re-prompting to force MFA set up, when used during manual testing from the host computer).

forevermatt (approved):

development/sp-local/config/authsources-pwmanager.php
briskt: It was strange that it even had an error. (Why only this one?) I could add it back later and do more diagnostics. Just wanted to keep moving forward.

docker-compose.yml
briskt: Actually, it wasn't solving a problem. I removed them after realizing that the test scripts never run composer in the `test` container.

PR #246, opened 2024-07-03
instantiate SimpleSAML\Utils classes

briskt:
### Changed
- Instantiate SimpleSAML\Utils classes. In SimpleSAMLphp 2, the Utils class methods have been made non-static.

https://itse.youtrack.cloud/issue/IDP-1103

[Reference](https://simplesamlphp.org/docs/stable/simplesamlphp-upgrade-notes-2.0.html#changes-relevant-for-module-developers)

### DIFF
--- a/modules/sildisco/public/metadata.php
+++ b/modules/sildisco/public/metadata.php
@@ -6,10 +6,7 @@
 require_once('../public/_include.php');
 
 use SAML2\Constants;
-use SimpleSAML\Utils\Auth as Auth;
-use SimpleSAML\Utils\Crypto as Crypto;
-use SimpleSAML\Utils\HTTP as HTTP;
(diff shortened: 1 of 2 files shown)


modules/sildisco/public/metadata.php
```diff
@@ -157,11 +154,13 @@
         $metaArray['scope'] = $idpmeta->getArray('scope');
     }
 
+    $metadataUtils = Utils\Metadata();
```
hobbitronics, line 157: Why no `new`? Is it just a method?

hobbitronics (approved):
Looks good, but have 1 question

modules/sildisco/public/metadata.php
briskt: Probably because I was tired!

PR #247, opened 2024-07-08
define variables needed in the header template

briskt:
### Fixed
- Define template variables needed in the header template (`theme_color_scheme` and `analytics_tracking_id`)

_Note: It may be possible to reduce this code duplication by defining our own Template class, inherited from `SimpleSAML\XHTML\Template`, but I think that even that would have to be duplicated once for each module. Perhaps it could be extracted to our [ssp-utilities](https://github.com/silinternational/ssp-utilities) repo._

### DIFF
--- a/modules/expirychecker/public/about2expire.php
+++ b/modules/expirychecker/public/about2expire.php
@@ -55,6 +55,8 @@ if (array_key_exists('changepwd', $_REQUEST)) {
 $globalConfig = Configuration::getInstance();
 
 $t = new Template($globalConfig, 'expirychecker:about2expire');
+$t->data['theme_color_scheme'] = $globalConfig->getOptionalString('theme.color-scheme', '');
+$t->data['analytics_tracking_id'] = $globalConfig->getOptionalString('analytics.trackingId', '');
 $t->data['form_target'] = Module::getModuleURL('expirychecker/about2expire.php');
 $t->data['form_data'] = ['StateId' => $stateId];
 $t->data['days_left'] = $state['daysLeft'];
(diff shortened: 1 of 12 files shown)


mtompset:
> _Note: It may be possible to reduce this code duplication by defining our own Template class, inherited from `SimpleSAML\XHTML\Template`,

Exactly what I thought after seeing all these tweaks.


> but I think that even that would have to be duplicated once for each module.

Duplicated how? As far as I can tell, the use line would change, that's it.


> Perhaps it could be extracted to our [ssp-utilities](https://github.com/silinternational/ssp-utilities) repo.

So, instead of changing the use line, add a functionality to do the tweaks on the template object? I think the extended object idea is better.

mtompset (approved):
The advantage of the extended object idea is that you don't accidentally forget to add these two lines everywhere. I'll approved, but it isn't the cleanest way. Maintenance headaches await.

briskt:
> > _Note: It may be possible to reduce this code duplication by defining our own Template class, inherited from `SimpleSAML\XHTML\Template`,
> 
> Exactly what I thought after seeing all these tweaks.
> 
> > but I think that even that would have to be duplicated once for each module.
> 
> Duplicated how? As far as I can tell, the use line would change, that's it.
> 
> > Perhaps it could be extracted to our [ssp-utilities](https://github.com/silinternational/ssp-utilities) repo.
> 
> So, instead of changing the use line, add a functionality to do the tweaks on the template object? I think the extended object idea is better.

There's a hard boundary between modules, such that you can't call code from one module to the other. It may be possible to do it in this repo, but outside of a module, but my PHP isn't strong enough to think of all the required details. I'd welcome anyone to step in and clean this up for me.

briskt:
> The advantage of the extended object idea is that you don't accidentally forget to add these two lines everywhere. I'll approved, but it isn't the cleanest way. Maintenance headaches await.

I note you said "I'll approve" but yet you didn't. Change of heart?

briskt:
> > The advantage of the extended object idea is that you don't accidentally forget to add these two lines everywhere. I'll approved, but it isn't the cleanest way. Maintenance headaches await.
> 
> I note you said "I'll approve" but yet you didn't. Change of heart?

Doh, my mistake. You did.

PR #249, opened 2024-07-10
Release 10.0.0-alpha.5 -- more fixes for SimpleSAMLphp 2

briskt:
### Changed
- Update Twig templates to recommended style (space after colon, snake case vars)

### Fixed
- Changed `getString` to `getOptionalString` (and similar changes)
- Instantiate SimpleSAMLphp Utils classes
- Define variables needed in the header template (`analytics_tracking_id` and `theme_color_scheme`)


modules/sildisco/public/sp/saml2-logout.php
```diff
@@ -58,8 +58,9 @@
 
 \SimpleSAML\Module\saml\Message::validateMessage($idpMetadata, $spMetadata, $message);
 
+$httpUtils = new \SimpleSAML\Utils\HTTP();
```
mtompset, line 61: I really like to import with use statements to keep dependencies more visible at the top of the file.

modules/sildisco/public/metadata.php
```diff
@@ -189,16 +188,16 @@
...
         $contacts = $idpmeta->getArray('contacts');
         foreach ($contacts as $contact) {
-            $metaArray['contacts'][] = Metadata::getContact($contact);
+            $metaArray['contacts'][] = $metadataUtils->getContact($contact);
         }
     }
 
-    $technicalContactEmail = $config->getString('technicalcontact_email', false);
-    if ($technicalContactEmail && $technicalContactEmail !== 'na@example.org') {
+    $technicalContactEmail = $config->getOptionalString('technicalcontact_email', null);
```
mtompset, line 195: Just curious, is there is a default of null, such that the second parameter is overkill?

mtompset (approved):
The whole swap to snake case made this extra long. Hopefully there are no values missing in the midst of all those cuts and changes.

modules/sildisco/public/sp/saml2-logout.php
briskt: Good suggestion. I'll make a note to do that across the board.

modules/sildisco/public/metadata.php
briskt: Good question. The signature is:

```php
public function getOptionalString(string $name, ?string $default): ?string
```

I think the second parameter is required.

briskt:
> The whole swap to snake case made this extra long. Hopefully there are no values missing in the midst of all those cuts and changes.

You're right. Fortunately, since the check for undefined template variables is enabled, and we have good test coverage, I think we're safe.

PR #250, opened 2024-07-10
Use more use statements

briskt:
### Fixed
- Added use statements to import classes
- Removed unnecessary use statements

### DIFF
--- a/dockerbuild/config/config.php
+++ b/dockerbuild/config/config.php
@@ -8,8 +8,9 @@
 
 use Sil\PhpEnv\Env;
 use Sil\PhpEnv\EnvVarNotFoundException;
+use SimpleSAML\Utils;
 
-$httpUtils = new \SimpleSAML\Utils\HTTP();
+$httpUtils = new Utils\HTTP();
 
 /*
  * Get config settings from ENV vars or set defaults
--- a/modules/silauth/public/status.php
+++ b/modules/silauth/public/status.php
@@ -28,5 +28,5 @@ try {
         $t->getMessage(),
         $t->getCode()
     );
-    \http_response_code(500);
+    http_response_code(500);
 }
--- a/modules/sildisco/public/metadata.php
+++ b/modules/sildisco/public/metadata.php
@@ -6,14 +6,19 @@
 require_once('../public/_include.php');
 
 use SAML2\Constants;
+use SimpleSAML\Configuration;
+use SimpleSAML\Error;
+use SimpleSAML\Metadata\MetaDataStorageHandler;
+use SimpleSAML\Metadata\SAMLBuilder;
+use SimpleSAML\Metadata\Signer;
 use SimpleSAML\Utils;
 
 // load SimpleSAMLphp, configuration and metadata
-$config = \SimpleSAML\Configuration::getInstance();
-$metadata = \SimpleSAML\Metadata\MetaDataStorageHandler::getMetadataHandler();
+$config = Configuration::getInstance();
+$metadata = MetaDataStorageHandler::getMetadataHandler();
 
 if (!$config->getOptionalBoolean('enable.saml20-idp', false)) {
-    throw new \SimpleSAML\Error\Error('NOACCESS');
+    throw new Error\Error('NOACCESS');
 }
 
 // check if valid local session exists
@@ -145,7 +150,7 @@ try {
         );
 
         if (!$idpmeta->hasValue('OrganizationURL')) {
-            throw new \SimpleSAML\Error\Exception('If OrganizationName is set, OrganizationURL must also be set.');
+            throw new Error\Exception('If OrganizationName is set, OrganizationURL must also be set.');
         }
         $metaArray['OrganizationURL'] = $idpmeta->getLocalizedString('OrganizationURL');
     }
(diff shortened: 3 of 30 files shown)


dockerbuild/config/config.php
```diff
@@ -8,8 +8,9 @@
 
 use Sil\PhpEnv\Env;
 use Sil\PhpEnv\EnvVarNotFoundException;
+use SimpleSAML\Utils;
 
-$httpUtils = new \SimpleSAML\Utils\HTTP();
+$httpUtils = new Utils\HTTP();
```
mtompset, line 13: I was expecting the use to include the HTTP, but I guess if the other two have HTTP defined, this would make sense.

modules/silauth/public/status.php
```diff
@@ -28,5 +28,5 @@
         $t->getMessage(),
         $t->getCode()
     );
-    \http_response_code(500);
+    http_response_code(500);
```
mtompset, line 31: I would have expected a more Yii'ish way of doing the same thing, not something so direct.

dockerbuild/config/config.php
briskt: I followed the example that SimpleSAMLphp devs use. Plus, I like how it reads: `new Utils\HTTP` rather than `new HTTP`.

modules/silauth/public/status.php
briskt: Yeah, it is kinda stark, isn't it?

modules/sildisco/public/metadata.php
```diff
@@ -6,14 +6,19 @@
...
 
 // load SimpleSAMLphp, configuration and metadata
-$config = \SimpleSAML\Configuration::getInstance();
-$metadata = \SimpleSAML\Metadata\MetaDataStorageHandler::getMetadataHandler();
+$config = Configuration::getInstance();
+$metadata = MetaDataStorageHandler::getMetadataHandler();
 
 if (!$config->getOptionalBoolean('enable.saml20-idp', false)) {
-    throw new \SimpleSAML\Error\Error('NOACCESS');
+    throw new Error\Error('NOACCESS');
```
mtompset, line 21: Perhaps a use ...\Error\Error as SimpleSAML_Error?

modules/sildisco/public/metadata.php
```diff
@@ -145,7 +150,7 @@
         );
 
         if (!$idpmeta->hasValue('OrganizationURL')) {
-            throw new \SimpleSAML\Error\Exception('If OrganizationName is set, OrganizationURL must also be set.');
+            throw new Error\Exception('If OrganizationName is set, OrganizationURL must also be set.');
```
mtompset, line 153: As above, but use ... as SimpleSAML_Exception?

modules/sildisco/public/metadata.php
```diff
@@ -224,5 +229,5 @@
         exit(0);
     }
 } catch (Exception $exception) {
-    throw new \SimpleSAML\Error\Error('METADATA', $exception);
+    throw new Error\Error('METADATA', $exception);
```
mtompset, line 232: I understand the less 'use' statements, but SimpleSAML_Error would be good in my opinion.

modules/sildisco/src/Auth/Process/AddIdp2NameId.php
```diff
@@ -124,7 +127,7 @@ public function process(array &$state): void
         $samlIDP = $state[self::IDP_KEY];
 
         if (empty($state[self::SP_NAMEID_ATTR])) {
-            \SimpleSAML\Logger::warning(
+            Logger::warning(
```
mtompset, line 130: "use ... as SS_Logger" would help distinguish the Logger, given the multitude of Logger's out there.

modules/sildisco/src/Auth/Process/AddIdp2NameId.php
```diff
@@ -138,25 +141,25 @@ public function process(array &$state): void
...
         }
-        $idpEntries = \Sil\SspUtils\Metadata::getIdpMetadataEntries($metadataPath);
+        $idpEntries = Metadata::getIdpMetadataEntries($metadataPath);
 
         $idpEntry = $idpEntries[$samlIDP];
 
         // The IDP metadata must have an IDPNamespace entry
         if (!isset($idpEntry[self::IDP_CODE_KEY])) {
-            throw new \SimpleSAML\Error\Exception(self::ERROR_PREFIX . "Missing required metadata entry: " .
+            throw new Error\Exception(self::ERROR_PREFIX . "Missing required metadata entry: " .
```
mtompset, line 150: Already commented about the Error\something above multiple times. Not going to comment on each one.

modules/sildisco/src/Auth/Process/TrackIdps.php
```diff
@@ -17,7 +20,7 @@ class TrackIdps extends \SimpleSAML\Auth\ProcessingFilter
     public function process(array &$state): void
     {
         // get the authenticating Idp and add it to the list of previous ones
-        $session = \SimpleSAML\Session::getSessionFromRequest();
+        $session = Session::getSessionFromRequest();
```
mtompset, line 23: Thinking about Session... perhaps a "use ... as SS_Session" might be useful, given there is a Yii session object.

mtompset (approved):
My comments are suggestions, not necessarily forced request for changes. This is already a nice improvement.

briskt:
@mtompset thanks for the alias suggestions. They might be a next step if/when things get messy or confusing.

PR #251, opened 2024-07-10
remove "beta test" feature

briskt:
### Removed
- Removed "beta test" feature. We do not use this in staging or production.


mtompset (approved):
Less is more. :)

jason-jackson (approved):
Looks good. I can't think of a reason why we would need it.

PR #252, opened 2024-07-10
misc. cleanup

briskt:
### Removed
- Removed leftover dictionary files from pre-SSP2
- Removed actions-services.yml
- Removed patch from run.sh


dockerbuild/run.sh
```diff
@@ -6,9 +6,6 @@ set -x
 # exit if any command fails
 set -e
 
-# This is a temporary fix (bug workaround) until ssp 2.0 is in use
```
mtompset, line 9: So, ssp 2.0 is now in use?

PR #253, opened 2024-07-11
Customized multi-IdP behavior

briskt:
### Changed
- Patch the standard SP AuthSource to modify behavior when multiple IdPs are available for an SP.
  - When more than one IdP is available, always present the discovery page. Standard behavior is to proceed using an existing IdP session if possible.
  - When re-authentication is required, always present the discovery page. Standard behavior is to let the user choose whether to cancel or continue.

### Fixed
- Fixed the SingleLogoutService on the test IdPs to use the standard saml2-logout.php since it no longer needs to be customized.

_**Note:** these customizations were done differently in ssp-base v9 (pre-SSP2). See sections marked "GTIS" in [this file](https://github.com/silinternational/simplesamlphp-module-sildisco/blob/develop/lib/Auth/Source/SP.php) for reference._


dockerbuild/ssp-overrides/sp-php.patch
jason-jackson, line 1: How difficult would this be maintain? If SimpleSamlPHP upgrades touch the underlying file, will it break this?

briskt: > How difficult would this be maintain? If SimpleSamlPHP upgrades touch the underlying file, will it break this?

If SimpleSAMLphp upgrades touch the lines in the patch range, it would break. But this method is far more stable than the previous method, in my opinion.

forevermatt: @briskt - I'm trying to see a diff of the before and after for this file, but when I checked out this branch, did a fresh build, and copied the SP.php file out of that docker container, it matches the version in simplesamlphp's GitHub repo. Can you help me with this?

forevermatt: Talked through this. I think I had copied the file out of an existing container, rather than killing it and starting it anew.

forevermatt (approved):

dockerbuild/ssp-overrides/sp-php.patch
```diff
@@ -1,16 +1,39 @@
...
+          if (
+              isset($state['saml:IDPList'])
+***************
+*** 858,866 ****
+                  $state['core:SP']
+              ));
+  
+!             $state['saml:sp:IdPMetadata'] = $this->getIdPMetadata($state['saml:sp:IdP']);
+!             $state['saml:sp:AuthId'] = $this->authId;
+!             self::askForIdPChange($state);
```
forevermatt, line 27: Does the `!` indicate a conflict of some sort? I see it used to be `<` and `>`, but now it's `!`.

briskt: No, it just means "changed line". Not sure if my commit message was clear, but from what I found today, this patch format is recommended because it includes context. Before this change, I noticed that the first "chunk" could be inserted anywhere in the file, which seemed dangerous to me.

forevermatt: Your commit message does make sense, thanks. I'm still a little confused why the patch-with-context stopped showing before-and-after for that second chunk, and now merely shows the result of the change. That's all, thanks.

briskt: No problem. I'm still learning the details of this myself. The "before" is still there. It's the portion after the `*** 858,866 ****` line, whereas the "after" lines follow the `--- 866,873 ----` line.

forevermatt: Ahh, okay. Thanks.

PR #254, opened 2024-07-16
fix the AddIdp2NameId filter

briskt:
### Fixed
- Fixed the AddIdp2NameId filter and sildisco unit tests

### DIFF
--- a/modules/sildisco/tests/AddIdpTest.php
+++ b/modules/sildisco/tests/AddIdpTest.php
@@ -14,7 +14,7 @@ class AddIdpTest extends TestCase
             'saml:sp:IdP' => $idp,
             'saml:sp:NameID' => [
                 [
-                    'Format' => 'urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified',
+                    'Format' => 'urn:oasis:names:tc:SAML:2.0:nameid-format:persistent',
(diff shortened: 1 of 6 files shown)


modules/sildisco/tests/AddIdpTest.php
```diff
@@ -105,23 +109,24 @@ public function testAddIdp2NameId_GoodString()
      */
     public function testAddIdp2NameId_GoodArray()
     {
+        $nameID = new NameID();
+        $nameID->setValue('Tester1_SmithA');
+        $nameID->setFormat('urn:oasis:names:tc:SAML:1.1:nameid-format:transient');
```
forevermatt, line 114: In other places, you change the format from SAML 1.1 transient to SAML 2.0 persistent. Do you want/need to do that here as well?

forevermatt (commented):
One question added, but overall it looks good.

modules/sildisco/tests/AddIdpTest.php
briskt: I will change it for consistency. Somewhere in the upgrade process, a change was made that seems to cause the format to be forced to 2.0 persistent. So the value set at the beginning of this test seems to be ignored.

forevermatt (approved):

PR #255, opened 2024-07-16
set global twig variables using controller

jason-jackson:
### Changed
- Use a controller to set global Twig variables

https://simplesamlphp.org/docs/stable/simplesamlphp-theming.html#a-custom-theme-controller


briskt (approved):
looks good!

PR #256, opened 2024-07-16
set trusted.url.domains using an environment variable

briskt:
### Added
- Added a new environment variable to set `trusted.url.domains`

https://itse.youtrack.cloud/issue/IDP-1125


briskt:
Requires https://github.com/silinternational/php-env/pull/26

hobbitronics (approved):
Looks good to me

PR #258, opened 2024-07-17
clean up logged error messages

briskt:
### Fixed
- Added a null coalesce on `$state['mfaLearnMoreUrl']` for the review page
  - fixes error: SimpleSAML\Error\Exception: Warning - Undefined array key "mfaLearnMoreUrl" at /data/vendor/simplesamlphp/simplesamlphp/modules/profilereview/public/nag.php:37
- Defined `exampleauth:UserPass` users in `users` key
  - fixes error: Module exampleauth:UserPass configured in legacy mode. Please put your username:password entries under the "users" key in your authsource.
- Changed from `if` to `assert` to catch error condition earlier
- Stop using the admin module to access the hub discovery page
  - fixes error: Caused by: Twig\Error\RuntimeError: An exception has been thrown during the rendering of a template ("SimpleSAML\XHTML\Template::getEntityDisplayName(): Argument #1 ($data) must be of type array, null given, called in /data/vendor/twig/twig/src/Environment.php(392) : eval()'d code on line 82").
- Filled in missing data in authsources.php
  - fixes error: SimpleSAML\Error\Exception: Warning - Undefined array key "add" at /data/vendor/simplesamlphp/simplesamlphp/modules/profilereview/src/Auth/Process/ProfileReview.php:234

### DIFF
--- a/modules/sildisco/src/IdPDisco.php
+++ b/modules/sildisco/src/IdPDisco.php
@@ -53,10 +53,8 @@ class IdPDisco extends SSPIdPDisco
         // Before the SimpleSAMLphp 2 upgrade, we added it to the state ourselves by overriding the SAML2.php file
         parse_str(parse_url($_GET['return'], PHP_URL_QUERY), $returnState);
         $state = Auth\State::loadState($returnState['AuthID'], 'saml:sp:sso');
-        if ($state && array_key_exists('SPMetadata', $state)) {
-            $spmd = $state['SPMetadata'];
-            $this->log('Updated SP metadata from ' . $this->spEntityId . ' to ' . $spmd['entityid']);
-        }
+        assert($state && array_key_exists('SPMetadata', $state));
+        $spmd = $state['SPMetadata'];
         $spEntityId = $spmd['entityid'];
 
         if (!empty($spEntityId)) {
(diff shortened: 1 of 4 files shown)


hobbitronics (approved):
Looks good to me

modules/sildisco/src/IdPDisco.php
```diff
@@ -53,10 +53,8 @@ private function getSPEntityIDAndReducedIdpList(): array
         // Before the SimpleSAMLphp 2 upgrade, we added it to the state ourselves by overriding the SAML2.php file
         parse_str(parse_url($_GET['return'], PHP_URL_QUERY), $returnState);
         $state = Auth\State::loadState($returnState['AuthID'], 'saml:sp:sso');
-        if ($state && array_key_exists('SPMetadata', $state)) {
-            $spmd = $state['SPMetadata'];
-            $this->log('Updated SP metadata from ' . $this->spEntityId . ' to ' . $spmd['entityid']);
-        }
+        assert($state && array_key_exists('SPMetadata', $state));
```
forevermatt, line 56: I don't think we want to depend on PHP's built in `assert()` function, since it is typically only run in debug or test mode, and is disabled in production mode (per [the docs](https://www.php.net/assert)).

If we want to ensure something is asserted every time, it seems to be better to use other/normal PHP code (either an `if` and `throw new Exception`, or  something like WebMozart's "Assert" library).

briskt: Oh, I didn't realize that. Thanks.

PR #259, opened 2024-07-17
remove the feature to disable an IdP

briskt:
### Removed
- Removed support for the `enabled` metadata IdP attribute


hobbitronics (approved):
lgtm

PR #261, opened 2024-07-18
remove logout template

briskt:
### Removed
- Removed material logout template. We have no need of this since logout redirects back to the SP.


hobbitronics (approved):
I guess there is no translation file for this? I did a search for material:logout:title but didn't see it in material.po

briskt:
> I guess there is no translation file for this? I did a search for material:logout:title but didn't see it in material.po

Good question. I had already deleted the old json translation file, but hadn't added them to the existing material.po files. (The `material:` prefix wouldn't be part of the key in material.po.)

PR #263, opened 2024-07-18
put the Dockerfile on a weight loss program

briskt:
### Removed
- In Dockerfile:
  - Removed ENV REFRESHED_AT
  - Removed RUN mkdir -p /data
  - Removed s3-expand
  - Removed RUN chmod a+x /data/run.sh /data/run-tests.sh
- Removed index.php override, originally added to call autoload.php, no longer needed
- Removed id.php override, originally added for load balancer, no longer needed
- Removed announcement.php placeholder

### Fixed
- Use preferred `ENV=value` syntax in Dockerfile


Dockerfile
```diff
@@ -52,20 +43,12 @@ COPY modules/material/themes/material/profilereview/* $SSP_PATH/modules/profiler
...
-COPY dockerbuild/ssp-overrides/id.php $SSP_PATH/public/id.php
-COPY dockerbuild/ssp-overrides/announcement.php $SSP_PATH/announcement/announcement.php
 COPY dockerbuild/ssp-overrides/sp-php.patch sp-php.patch
 RUN patch /data/vendor/simplesamlphp/simplesamlphp/modules/saml/src/Auth/Source/SP.php sp-php.patch
 
-COPY tests /data/tests
-
-RUN chmod a+x /data/run.sh /data/run-tests.sh
-
 ADD https://github.com/silinternational/config-shim/releases/latest/download/config-shim.gz config-shim.gz
```
jason-jackson, line 52: It might be worth providing a non-gziped version of the config-shim so we can do something like:

```ADD --chmod=755 https://github.com/silinternational/config-shim/releases/latest/download/config-shim /usr/local/bin/config-shim```

See: https://docs.docker.com/reference/dockerfile/#add---chown---chmod

Then actual questions:

- What gets updated more - the above lines or this? (Meaning, should this ADD be higher up, so it doesn't need to be rewritten with as much frequency) I think this would be updated less, but both are probably fairly infrequent.
- Do we need to set this to a specific version (instead of `latest` so Docker will see it needs to updated if we update it?

dockerbuild/ssp-overrides/announcement.php
jason-jackson, line 1: Are we still doing these overrides? Or will that be done more on IdP or hub level?

jason-jackson (approved):
Looks good. Just minor comments

Dockerfile
briskt: > ...non-gziped version...
Since Go binaries are not small, this could have negative ramifications. I think the extra line in Dockerfile is worth it.

> What gets updated more...
Hard to say. I didn't give much consideration to this, though I agree it's a good way to optimize.

> ...specific version...
Probably so.

dockerbuild/ssp-overrides/announcement.php
briskt: Yes, we are still using announcement.php and technically it's not an "override" since stock SSP doesn't have this feature. The one being removed in this PR is effectively an empty file, so it seemed pointless to have it. The code that looks for the file checks for its existence before including it.

PR #265, opened 2024-07-18
Release 10.0.0-alpha.7 -- miscellaneous fixes

briskt:
### Fixed
- Use preferred `ENV=value` syntax in Dockerfile
- Correct the Metadata utils path in sildisco/metadata.php
- Added a null coalesce on `$state['mfaLearnMoreUrl']` for the review page
  - fixes error: SimpleSAML\Error\Exception: Warning - Undefined array key "mfaLearnMoreUrl" at /data/vendor/simplesamlphp/simplesamlphp/modules/profilereview/public/nag.php:37
- Defined `exampleauth:UserPass` users in `users` key
  - fixes error: Module exampleauth:UserPass configured in legacy mode. Please put your username:password entries under the "users" key in your authsource.
- Changed from `if` to `assert` to catch error condition earlier
- Stop using the admin module to access the hub discovery page
  - fixes error: Caused by: Twig\Error\RuntimeError: An exception has been thrown during the rendering of a template ("SimpleSAML\XHTML\Template::getEntityDisplayName(): Argument #⁠1 ($data) must be of type array, null given, called in /data/vendor/twig/twig/src/Environment.php(392) : eval()'d code on line 82").
- Filled in missing data in authsources.php
  - fixes error: SimpleSAML\Error\Exception: Warning - Undefined array key "add" at /data/vendor/simplesamlphp/simplesamlphp/modules/profilereview/src/Auth/Process/ProfileReview.php:234

### Removed
- Removed browser check for webauthn support
- Removed abandoned package sinergi/browser-detector
- In Dockerfile:
  - Removed ENV REFRESHED_AT
  - Removed RUN mkdir -p /data
  - Removed s3-expand
  - Removed RUN chmod a+x /data/run.sh /data/run-tests.sh
- Removed index.php override, originally added to call autoload.php, no longer needed
- Removed id.php override, originally added for load balancer, no longer needed
- Removed announcement.php placeholder
- Removed material logout template. We have no need of this since logout redirects back to the SP.
- Removed support for the enabled metadata IdP attribute

### Changed
- Updated php-env and use `null` as the default for `getArray` for TRUSTED_URL_DOMAINS


modules/material/themes/material/default/selectidp-links.twig
```diff
@@ -78,36 +78,9 @@
...
-              </span>
-            </div>
-          </div>
-        </div>
-      {% endfor %}
     </form>
   </main>
 
-  <script>
-    ga('send', 'event', 'hub', 'choices', 'enabled', {{ enabled_idps|length }});
```
jason-jackson, line 107: Do we still want to send this event off? (with idp_list)

jason-jackson (approved):

modules/material/themes/material/default/selectidp-links.twig
briskt: Seems like pretty useless information to me. GA makes more sense for discovering user behavior. But it's all useless because nobody has ever used the data we collect anyway.

PR #268, opened 2024-07-19
remove config options that are not used

briskt:
### Removed
- Removed `originalUrlParam`
- Removed `excludeByDefault`
- Removed `IDP_NAME` and `IDP_DISPLAY_NAME` and use the metadata `name` property instead
- Removed `SAML20_IDP_ENABLE`
- Removed `SESSION_DURATION`

_**Note:** to study each removal separately, look at each commit one-by-one._


forevermatt (approved):
Yeah for cutting stuff we don't need any more 😁

PR #269, opened 2024-07-22
throw exception if SP entityID can't be found in the state

briskt:
### Fixed
- Rather than fail silently, throw an exception if the state doesn't contain the SP entityID

### DIFF
--- a/modules/sildisco/src/IdPDisco.php
+++ b/modules/sildisco/src/IdPDisco.php
@@ -6,10 +6,12 @@ use Sil\SspUtils\AnnouncementUtils;
 use Sil\SspUtils\DiscoUtils;
 use Sil\SspUtils\Metadata;
 use SimpleSAML\Auth;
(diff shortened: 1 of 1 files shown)


modules/sildisco/src/IdPDisco.php
```diff
@@ -39,6 +41,10 @@ private function getMetadataPath()
         return __DIR__ . '/../../../metadata/';
     }
 
+    /**
+     * @throws NoState
```
forevermatt, line 45: I don't see where this `NoState` error is actually being thrown. Did you intend to throw that on lines 60 and/or 65?

forevermatt (approved):

modules/sildisco/src/IdPDisco.php
briskt: PHPStorm added that. It's in `Auth\State::loadState` which is why I removed the check on `$state`.

PR #271, opened 2024-07-23
Feature/test cleanup

jason-jackson:
### Changed
- Fix announcement tests
- Add description of all metadata tests to README
- Cleanup metadata tests to use corresponding assert

### DIFF
--- a/tests/AnnouncementTest.php
+++ b/tests/AnnouncementTest.php
@@ -15,8 +15,8 @@ class AnnouncementTest extends TestCase
     {
         $announcementPathFile = '/data/ssp-announcement.php';
         if (file_exists($announcementPathFile)) {
-            $announcement = include $announcementPathFile;
-            $this->assertIsString($announcement);
+            $results = AnnouncementUtils::getSimpleAnnouncement();
+            $this->assertNotNull($results);
         }
     }
 
--- a/tests/AnnouncementTest.php
+++ b/tests/AnnouncementTest.php
@@ -13,7 +13,11 @@ class AnnouncementTest extends TestCase
      */
     public function testGetSimpleAnnouncement()
     {
-        $results = AnnouncementUtils::getSimpleAnnouncement();
+        $announcementPathFile = '/data/ssp-announcement.php';
+        if (file_exists($announcementPathFile)) {
+            $announcement = include $announcementPathFile;
+            $this->assertIsString($announcement);
+        }
     }
 
 }
--- a/tests/MetadataTest.php
+++ b/tests/MetadataTest.php
@@ -55,13 +55,13 @@ class MetadataTest extends TestCase
         $spFiles = $this->getSpMetadataFiles();
         foreach ($spFiles as $file) {
             $returnVal = include $file;
-            $this->assertTrue(is_array($returnVal), 'Metadata file does not return array as expected. File: ' . $file);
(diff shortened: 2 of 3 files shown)


briskt (approved):

tests/MetadataTest.php
```diff
@@ -55,13 +55,13 @@ public function testMetadataFilesReturnArrays()
...
             $returnVal = include $file;
-            $this->assertTrue(is_array($returnVal), 'Metadata file does not return array as expected. File: ' . $file);
+            $this->assertIsArray($returnVal, 'Metadata file does not return array as expected. File: ' . $file);
         }
 
         $idpFiles = $this->getIdPMetadataFiles();
         foreach ($idpFiles as $file) {
             $returnVal = include $file;
-            $this->assertTrue(is_array($returnVal), 'Metadata file does not return array as expected. File: ' . $file);
+            $this->assertIsArray($returnVal, 'Metadata file does not return array as expected. File: ' . $file);
```
mtompset, line 64: If you look at the assertions for PHPUnit, you'll notice they are static. Have the same clean up mess in MailAdmin to deal with.

tests/AnnouncementTest.php
```diff
@@ -13,7 +13,11 @@ class AnnouncementTest extends TestCase
      */
     public function testGetSimpleAnnouncement()
     {
-        $results = AnnouncementUtils::getSimpleAnnouncement();
+        $announcementPathFile = '/data/ssp-announcement.php';
```
mtompset, line 16: Why this change?

mtompset (changes requested):
Nice cleanup of lazy assertTrue's and assertFalse's to more proper assertions. Though, explain that one change before merging, please.

tests/AnnouncementTest.php
briskt: Actually, that's a good question. This test doesn't call the getSimpleAnnouncement function now. Shouldn't it?

jason-jackson: The problem is that `getSimpleAnnouncement` won't ever fail or throw an exception, instead it returns null.

So this now checks the file that that function uses and sees if it works. I guess I could do this instead:
```
        $announcementPathFile = '/data/ssp-announcement.php';
        if (file_exists($announcementPathFile)) {
            $results = AnnouncementUtils::getSimpleAnnouncement();
            $this->assertNotNull($results);
        }
```

which is actually probably a better idea.

tests/MetadataTest.php
jason-jackson: > The assertion methods are declared static and can be invoked from any context using PHPUnit\Framework\Assert::assertTrue(), for instance, or using $this->assertTrue() or self::assertTrue(), for instance, in a class that extends PHPUnit\Framework\TestCase. You can even use global function wrappers such as assertTrue().

https://docs.phpunit.de/en/11.2/assertions.html#static-vs-non-static-usage-of-assertion-methods

We use `class AnnouncementTest extends TestCase` , so it is fine as it is.

tests/AnnouncementTest.php
jason-jackson: I figure we want to have it fail if we try to add an announcement, but it doesn't work; however, if no announcement file exists, we want to allow that through

briskt (approved):

briskt (approved):

PR #275, opened 2024-08-05
remove unused code

briskt:


modules/expirychecker/src/Utilities.php
```diff
@@ -54,7 +12,7 @@ public static function haveSameDomain(
      **/
     public static function getUrlFromRelayState(string $relayState): string
     {
-        if (strpos($relayState, "http") === 0) {
+        if (str_starts_with($relayState, "http")) {
```
mtompset, line 15: Prettier!

modules/silauth/src/Auth/Source/auth/IdBroker.php
```diff
@@ -95,16 +95,4 @@ public function getAuthenticatedUser(string $username, string $password): ?array
...
     }
-
-    /**
-     * Ping the /site/status URL. If the ID Broker's status is fine, the
-     * response string is returned. If not, an exception is thrown.
-     *
-     * @return string "OK"
-     * @throws Exception
-     */
-    public function getSiteStatus(): string
```
mtompset, line 106: Why is this being removed? Is there no more pinging?

modules/expirychecker/src/Utilities.php
briskt: I appreciate IDE suggestions like this. Honestly don't know how people get by with simple editors.

modules/silauth/src/Auth/Source/auth/IdBroker.php
briskt: Yes, but since ID Broker has its own status check pinged directly by ECS and NodePing, we removed the check done from the ssp container a while back.

hobbitronics (approved):
Looks good

jason-jackson (approved):

modules/silauth/src/Auth/Source/auth/IdBroker.php
mtompset: What function does NodePing trigger?

modules/expirychecker/src/Utilities.php
mtompset: The difficulty is that is a PHP 8.x function. Some people haven't transitioned from 7.x yet.

modules/silauth/src/Auth/Source/auth/IdBroker.php
briskt: https://github.com/silinternational/ssp-base/blob/develop/modules/silauth/public/status.php

modules/expirychecker/src/Utilities.php
briskt: This isn't a library though, so it doesn't matter. Good to know, though, in case I encounter that in library code.

mtompset (approved):

Issue #203, opened 2024-05-27
nag.php expects array entries which aren't always being set


briskt:
We either need to update www/nag.php to handle missing entries for methodOptions, mfaOptions, and mfaLearnMoreUrl, or we need to update lib/auth/process/ProfileReview.php to always provide values for those in the $state whenever sending the user to that nag.php page.

(Originally filed in the [profilereview repo](https://github.com/silinternational/simplesamlphp-module-profilereview/issues/26).)

PR #278, opened 2024-09-16
Release 10.0.2 -- assign all values nag controller needs

briskt:
[IDP-1043](https://itse.youtrack.cloud/issue/IDP-1043) ssp-base #203: nag.php expects array entries which aren't always being set

### Fixed
- Assign all values needed in the state by the nag.php page.

### Changed
- Refactor ProfileReview.php `redirectToProfileReview` and `redirectToNag` methods to pass fewer parameters and reduce code duplication.


hobbitronics (approved):
Looks good

briskt:
Resolves #203

PR #281, opened 2024-09-20
Idp 781 update last_login_utc from inside mfa

hobbitronics:
[IDP-781](https://itse.youtrack.cloud/issue/IDP-781)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "ext-gmp": "*",
     "ext-json": "*",
     "codemix/yii2-streamlog": "^1.3",
-    "simplesamlphp/simplesamlphp": "^2.0",
+    "simplesamlphp/simplesamlphp": ">=2.2.2 <2.3.0",
     "simplesamlphp/composer-module-installer": "^1.0",
     "rlanvin/php-ip": "^1.0",
     "silinternational/ssp-utilities": "v2.0.0",
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     "silinternational/php-env": "^3.2.0",
     "silinternational/psr3-adapters": "v4.0.0",
     "silinternational/yii2-json-log-targets": "^2.0",
-    "silinternational/idp-id-broker-php-client": "^4.3",
+    "silinternational/idp-id-broker-php-client": "^4.4",
     "yiisoft/yii2": "~2.0.12",
     "yiisoft/yii2-gii": "^2.0",
     "google/recaptcha": "^1.1",
@@ -49,7 +49,8 @@
   "config": {
     "allow-plugins": {
       "simplesamlphp/composer-module-installer": true,
-      "yiisoft/yii2-composer": true
+      "yiisoft/yii2-composer": true,
+      "simplesamlphp/composer-xmlprovider-installer": true
     }
   }
 }
--- a/features/fakes/FakeIdBrokerClient.php
+++ b/features/fakes/FakeIdBrokerClient.php
@@ -57,6 +57,26 @@ class FakeIdBrokerClient
     }
 
     /**
+     * Update the User last_login_utc by calling the internal method
+     * @param string $employeeId
+     * @return array An array of information about the updated user only including last_login_utc and employee_id
+     * @throws Exception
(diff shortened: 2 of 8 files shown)


hobbitronics:
Woops, looks like this isn't quite ready yet.

features/fakes/FakeIdBrokerClient.php
```diff
@@ -56,6 +56,26 @@ public function mfaVerify(int $id, string $employeeId, string|array $value): arr
         ];
     }
 
+    /**
+     * Update the User last_login_utc by calling the internal method
+     * @param string $employeeId
+     * @return array An array of information about the updated user only including last_login_utc and employee_id
+     * @throws Exception
+     */
+    public function updateUserLastLogin(string $employeeId): array
```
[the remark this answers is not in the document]
hobbitronics: I'm not sure what is supposed to happen here.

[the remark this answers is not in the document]
briskt: I think you have it correct, @hobbitronics. The initial comment by @mtompset was made before he realized it was test-only code.

modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -161,7 +161,7 @@ protected function getAttributeAllValues(string $attributeName, array $state): ?
      * @param array $idBrokerConfig
      * @return IdBrokerClient
      */
-    protected static function getIdBrokerClient(array $idBrokerConfig): IdBrokerClient|FakeIdBrokerClient
+    static function getIdBrokerClient(array $idBrokerConfig): IdBrokerClient|FakeIdBrokerClient
```
hobbitronics, line 164:
```suggestion
    public static function getIdBrokerClient(array $idBrokerConfig): IdBrokerClient|FakeIdBrokerClient
```

modules/mfa/public/prompt-for-mfa.php
```diff
@@ -32,6 +32,8 @@
 $cookieHash = filter_input(INPUT_COOKIE, 'c1') ?? ''; // hashed string
 $expireDate = filter_input(INPUT_COOKIE, 'c2') ?? 0;  // expiration timestamp
 if (Mfa::isRememberMeCookieValid(base64_decode($cookieHash), $expireDate, $mfaOptions, $state)) {
+    $idBrokerClient = Mfa::getIdBrokerClient($state['idBrokerConfig']);
+    $idBrokerClient->updateUserLastLogin($state['employeeId']);
```
forevermatt, line 36: I think this sort of call should be happening in the AuthProc, not here in the view/controller file. Want to find a time to talk through this together?

hobbitronics: I'd actually thought about moving it back to mfa, I just hadn't gotten to it yet.

composer.json
```diff
@@ -49,7 +49,8 @@
   "config": {
     "allow-plugins": {
       "simplesamlphp/composer-module-installer": true,
-      "yiisoft/yii2-composer": true
+      "yiisoft/yii2-composer": true,
+      "simplesamlphp/composer-xmlprovider-installer": true
```
briskt, line 53: Why was this added? I'm guessing it was automatically added, but it can probably be removed or changed to `false` since we don't use it as far as I can tell.

modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -161,7 +161,7 @@ protected function getAttributeAllValues(string $attributeName, array $state): ?
      * @param array $idBrokerConfig
      * @return IdBrokerClient
      */
-    protected static function getIdBrokerClient(array $idBrokerConfig): IdBrokerClient|FakeIdBrokerClient
+    private static function getIdBrokerClient(array $idBrokerConfig): IdBrokerClient|FakeIdBrokerClient
```
briskt, line 164: Good to make this more restricted.

briskt (approved):

composer.json
hobbitronics: It was automatically updated.

PR #283, opened 2024-10-09
Release 10.1.4 -- resolve lint warnings

briskt:
### Fixed
- Resolve some of the lint warnings identified by PhpStorm

### DIFF
--- a/features/fakes/FakeIdBrokerClient.php
+++ b/features/fakes/FakeIdBrokerClient.php
@@ -2,6 +2,7 @@
 
 namespace Sil\SspBase\Features\fakes;
 
+use Exception;
 use InvalidArgumentException;
 use Sil\Idp\IdBroker\Client\ServiceException;
 
@@ -84,7 +85,7 @@ class FakeIdBrokerClient
      * @return array|null
      * @throws Exception
      */
-    public function mfaCreate($employee_id, $type, $label = null)
+    public function mfaCreate($employee_id, $type, $label = null): ?array
     {
         if (empty($employee_id)) {
             throw new InvalidArgumentException('employee_id is required');
@@ -125,9 +126,8 @@ class FakeIdBrokerClient
      * Get a list of MFA configurations for given user
      * @param string $employee_id
      * @return array
-     * @throws ServiceException
      */
-    public function mfaList($employee_id)
+    public function mfaList(string $employee_id)
     {
         return [
             [
--- a/features/fakes/FakeIdBrokerClient.php
+++ b/features/fakes/FakeIdBrokerClient.php
@@ -81,11 +81,11 @@ class FakeIdBrokerClient
      * Create a new MFA configuration
      * @param string $employee_id
      * @param string $type
-     * @param string $label
+     * @param string|null $label
      * @return array|null
      * @throws Exception
      */
-    public function mfaCreate($employee_id, $type, $label = null): ?array
+    public function mfaCreate(string $employee_id, string $type, string $label = null): ?array
     {
         if (empty($employee_id)) {
             throw new InvalidArgumentException('employee_id is required');
@@ -127,7 +127,7 @@ class FakeIdBrokerClient
      * @param string $employee_id
      * @return array
      */
-    public function mfaList(string $employee_id)
+    public function mfaList(string $employee_id): array
     {
         return [
             [
--- a/modules/expirychecker/src/Auth/Process/ExpiryDate.php
+++ b/modules/expirychecker/src/Auth/Process/ExpiryDate.php
@@ -37,6 +37,7 @@ class ExpiryDate extends ProcessingFilter
      *
      * @param array $config Configuration information about this filter.
      * @param mixed $reserved For future use.
+     * @throws Exception
      */
     public function __construct(array $config, mixed $reserved)
     {
@@ -65,6 +66,9 @@ class ExpiryDate extends ProcessingFilter
         ]);
     }
 
+    /**
+     * @throws Exception
+     */
     protected function loadValuesFromConfig(array $config, array $attributeRules): void
     {
         foreach ($attributeRules as $attribute => $rules) {
@@ -141,6 +145,9 @@ class ExpiryDate extends ProcessingFilter
         return $expiryTimestamp;
     }
 
+    /**
+     * @throws Exception
+     */
     public static function hasSeenSplashPageRecently(): bool
     {
         $session = Session::getSessionFromRequest();
@@ -150,6 +157,9 @@ class ExpiryDate extends ProcessingFilter
         );
     }
 
+    /**
+     * @throws Exception
+     */
     public static function skipSplashPagesFor(int $seconds): void
     {
         $session = Session::getSessionFromRequest();
@@ -162,6 +172,9 @@ class ExpiryDate extends ProcessingFilter
         $session->save();
     }
 
+    /**
+     * @throws Exception
+     */
     protected function initLogger(array $config): void
     {
         $loggerClass = $config['loggerClass'] ?? Psr3SamlLogger::class;
@@ -215,6 +228,7 @@ class ExpiryDate extends ProcessingFilter
 
     /**
      * @inheritDoc
+     * @throws Exception
      */
     public function process(array &$state): void
     {
@@ -244,7 +258,7 @@ class ExpiryDate extends ProcessingFilter
         ]));
 
         if ($this->isExpired($expiryTimestamp)) {
-            $this->redirectToExpiredPage($state, $accountName, $expiryTimestamp);
+            $this->redirectToExpiredPage($state, $accountName);
         }
 
         // Display a password expiration warning page if it's time to do so.
@@ -263,9 +277,8 @@ class ExpiryDate extends ProcessingFilter
      *
      * @param array $state The state data.
      * @param string $accountName The name of the user account.
-     * @param int $expiryTimestamp When the password expired.
      */
-    public function redirectToExpiredPage(array &$state, string $accountName, int $expiryTimestamp): void
+    public function redirectToExpiredPage(array &$state, string $accountName): void
     {
         assert('is_array($state)');
 
--- a/modules/mfa/public/prompt-for-mfa.php
+++ b/modules/mfa/public/prompt-for-mfa.php
@@ -68,7 +68,7 @@ $mfaOption = Mfa::getMfaOptionById($mfaOptions, $mfaId);
 if (filter_has_var(INPUT_POST, 'submitMfa')) {
     /* @var string|array $mfaSubmission */
     $mfaSubmission = filter_input(INPUT_POST, 'mfaSubmission');
-    if (substr($mfaSubmission, 0, 1) == '{') {
+    if (str_starts_with($mfaSubmission, '{')) {
         $mfaSubmission = json_decode($mfaSubmission, true);
     }
 
--- a/modules/mfa/public/send-manager-mfa.php
+++ b/modules/mfa/public/send-manager-mfa.php
@@ -23,6 +23,7 @@ $state = State::loadState($stateId, Mfa::STAGE_SENT_TO_MFA_PROMPT);
 
 $logger = LoggerFactory::getAccordingToState($state);
 
+$errorMessage = null;
 if (filter_has_var(INPUT_POST, 'send')) {
     try {
         Mfa::sendManagerCode($state, $logger);
@@ -41,7 +42,7 @@ $globalConfig = Configuration::getInstance();
 
 $t = new Template($globalConfig, 'mfa:send-manager-mfa');
 $t->data['manager_email'] = $state['managerEmail'];
-$t->data['error_message'] = $errorMessage ?? null;
+$t->data['error_message'] = $errorMessage;
 $t->send();
 
 $logger->info(json_encode([
--- a/modules/mfa/public/send-manager-mfa.php
+++ b/modules/mfa/public/send-manager-mfa.php
@@ -24,7 +24,11 @@ $state = State::loadState($stateId, Mfa::STAGE_SENT_TO_MFA_PROMPT);
 $logger = LoggerFactory::getAccordingToState($state);
 
 if (filter_has_var(INPUT_POST, 'send')) {
-    $errorMessage = Mfa::sendManagerCode($state, $logger);
+    try {
+        Mfa::sendManagerCode($state, $logger);
+    } catch (Exception $e) {
+        $errorMessage = $e->getMessage();
+    }
 } elseif (filter_has_var(INPUT_POST, 'cancel')) {
     $moduleUrl = SimpleSAML\Module::getModuleURL('mfa/prompt-for-mfa.php', [
         'StateId' => $stateId,
--- a/modules/mfa/src/Auth/Process/Mfa.php
+++ b/modules/mfa/src/Auth/Process/Mfa.php
@@ -17,6 +17,7 @@ use SimpleSAML\Auth\State;
 use SimpleSAML\Module;
 use SimpleSAML\Module\mfa\LoggerFactory;
 use SimpleSAML\Utils\HTTP;
+use Throwable;
 
 /**
  * Filter which prompts the user for MFA credentials.
@@ -25,9 +26,7 @@ use SimpleSAML\Utils\HTTP;
  */
 class Mfa extends ProcessingFilter
 {
-    const SESSION_TYPE = 'mfa';
     const STAGE_SENT_TO_LOW_ON_BACKUP_CODES_NAG = 'mfa:sent_to_low_on_backup_codes_nag';
-    const STAGE_SENT_TO_MFA_CHANGE_URL = 'mfa:sent_to_mfa_change_url';
     const STAGE_SENT_TO_MFA_NEEDED_MESSAGE = 'mfa:sent_to_mfa_needed_message';
     const STAGE_SENT_TO_MFA_PROMPT = 'mfa:sent_to_mfa_prompt';
     const STAGE_SENT_TO_NEW_BACKUP_CODES_PAGE = 'mfa:sent_to_new_backup_codes_page';
@@ -79,6 +78,9 @@ class Mfa extends ProcessingFilter
         $this->idBrokerClientClass = $config['idBrokerClientClass'] ?? IdBrokerClient::class;
     }
 
+    /**
+     * @throws Exception
+     */
     protected function loadValuesFromConfig(array $config, array $attributes): void
     {
         foreach ($attributes as $attribute) {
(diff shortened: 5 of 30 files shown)


features/fakes/FakeIdBrokerClient.php
```diff
@@ -84,7 +85,7 @@ public function updateUserLastLogin(string $employeeId): array
      * @return array|null
      * @throws Exception
      */
-    public function mfaCreate($employee_id, $type, $label = null)
+    public function mfaCreate($employee_id, $type, $label = null): ?array
```
mtompset, line 88: no strong typing on the parameter list? I realize that sometimes uglinesses exist and not typing is nicer. Just asking, because you strong typed below.

features/fakes/FakeIdBrokerClient.php
```diff
@@ -125,9 +126,8 @@ public function mfaCreate($employee_id, $type, $label = null)
      * Get a list of MFA configurations for given user
      * @param string $employee_id
      * @return array
-     * @throws ServiceException
      */
-    public function mfaList($employee_id)
+    public function mfaList(string $employee_id)
```
mtompset, line 130: Did you check all the calls to make sure everything was passing a defined value which was a string? Because debugging a type mismatch as a result of a function call is a pain to debug.

modules/expirychecker/src/Auth/Process/ExpiryDate.php
```diff
@@ -244,7 +258,7 @@ public function process(array &$state): void
         ]));
 
         if ($this->isExpired($expiryTimestamp)) {
-            $this->redirectToExpiredPage($state, $accountName, $expiryTimestamp);
+            $this->redirectToExpiredPage($state, $accountName);
```
mtompset, line 261: Why was the expiryTimestamp removed?

modules/mfa/public/prompt-for-mfa.php
```diff
@@ -68,7 +68,7 @@
 if (filter_has_var(INPUT_POST, 'submitMfa')) {
     /* @var string|array $mfaSubmission */
     $mfaSubmission = filter_input(INPUT_POST, 'mfaSubmission');
-    if (substr($mfaSubmission, 0, 1) == '{') {
+    if (str_starts_with($mfaSubmission, '{')) {
```
mtompset, line 71: Yay for newer versions of PHP with better string handling functions. :)

modules/mfa/public/send-manager-mfa.php
```diff
@@ -24,7 +24,11 @@
 $logger = LoggerFactory::getAccordingToState($state);
 
 if (filter_has_var(INPUT_POST, 'send')) {
-    $errorMessage = Mfa::sendManagerCode($state, $logger);
+    try {
+        Mfa::sendManagerCode($state, $logger);
+    } catch (Exception $e) {
+        $errorMessage = $e->getMessage();
```
mtompset, line 30: Using the ?? null on 44 is laziness. You could define errorMessage on 25 as null. I get the saving vertical space and one instruction slower issue, but explicitly defining would be better in my opinion.

modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -416,15 +418,16 @@ protected static function isHeadedToMfaSetupUrl($state, $mfaSetupUrl)
      *
      * @param int $mfaId The ID of the MFA option used.
      * @param string $employeeId The Employee ID that this MFA option belongs to.
-     * @param string $mfaSubmission The value of the MFA submission.
+     * @param string|array $mfaSubmission The value of the MFA submission.
```
mtompset, line 421: Is there a way to single data type the parameter? If it is a string, you can always call with [ $string ] to make it an array.

modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -416,15 +418,16 @@ protected static function isHeadedToMfaSetupUrl($state, $mfaSetupUrl)
...
      * @param array $state The array of state information.
      * @param bool $rememberMe Whether or not to set remember me cookies
      * @param LoggerInterface $logger A PSR-3 compatible logger.
      * @param string $mfaType The type of the MFA ('webauthn', 'totp', 'backupcode').
      * @param string $rpOrigin The Relying Party Origin (for WebAuthn)
      * @return string If validation fails, an error message to show to the
      *     end user will be returned.
-     * @throws \Sil\PhpEnv\EnvVarNotFoundException
+     * @throws EnvVarNotFoundException
+     * @throws Exception
```
mtompset, line 430: It's not letting me comment on the code 442-450. So this only needs one of those things defined? Because normally, I would expect all parameters to be provided for a function.

modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -539,7 +542,7 @@ public static function validateMfaSubmission(
      *
      * @param array $state
      */
-    public static function redirectToMfaSetup(array &$state): void
+    public static function redirectToMfaSetup(array $state): void
```
mtompset, line 545: Arrays are already pointers. You can change the contents if you pass them. :)

modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -818,8 +821,9 @@ protected static function shouldPromptForMfa(array $mfa): bool
      *
      * @param array $state The state data.
      * @param LoggerInterface $logger A PSR-3 compatible logger.
+     * @throws Exception
      */
-    public static function sendManagerCode(array &$state, LoggerInterface $logger): string
+    public static function sendManagerCode(array &$state, LoggerInterface $logger): void
```
mtompset, line 826: Yikes. That's weird it was string before.

modules/profilereview/src/Auth/Process/ProfileReview.php
```diff
@@ -85,7 +85,7 @@ protected function loadValuesFromConfig(array $config, array $attributes): void
      * @param LoggerInterface $logger The logger.
      * @throws Exception
      */
-    public static function validateConfigValue($attribute, $value, $logger)
+    public static function validateConfigValue(string $attribute, mixed $value, LoggerInterface $logger): void
```
mtompset, line 88: Have you checked that strong typing won't break any functions?

modules/profilereview/src/Auth/Process/ProfileReview.php
```diff
@@ -110,7 +110,7 @@ public static function validateConfigValue($attribute, $value, $logger)
      * @param array $state The state data.
      * @return mixed The attribute value, or null if not found.
      */
-    protected function getAttribute($attributeName, $state)
+    protected function getAttribute(string $attributeName, array $state): mixed
```
mtompset, line 113: Have you checked if this is an overriding function and that the typing matches the underlying library's definition of getAttribute? (generalize for all functions you are strong typing)

modules/profilereview/src/Auth/Process/ProfileReview.php
```diff
@@ -148,36 +148,36 @@ protected function getAttributeAllValues($attributeName, $state)
...
-            if (strpos($samlRelayState, "https://") === 0) {
+            if (str_starts_with($samlRelayState, "https://")) {
                 return $samlRelayState;
             }
         }
         return '';
     }
 
-    protected function initComposerAutoloader()
+    protected function initComposerAutoloader(): void
```
mtompset, line 167: That checking of a function's type definitions against a library's overridden function includes the return type.

modules/silauth/src/Auth/Source/auth/AuthError.php
```diff
@@ -17,7 +17,7 @@ class AuthError
     const CODE_RATE_LIMIT_MINUTES = 'rate_limit_minutes';
 
     private string $code;
-    private array $messageParams = [];
+    private array $messageParams;
```
mtompset, line 20: If you don't initialize and then you access before it is set, PHP throws an error.  Is there a nullable case?

modules/silauth/src/Auth/Source/time/UtcTime.php
```diff
@@ -4,14 +4,15 @@
...
+use Exception;
 use InvalidArgumentException;
 
 class UtcTime
 {
     const DATE_TIME_FORMAT = 'Y-m-d H:i:s';
 
-    private $utc;
-    private $dateTime;
+    private DateTimeZone $utc;
```
mtompset, line 14: Is this always initialized before used?

modules/silauth/src/Auth/Source/time/UtcTime.php
```diff
@@ -4,14 +4,15 @@
...
+    private DateTime $dateTime;
```
mtompset, line 15: Is this always initialized before used?

modules/silauth/src/Auth/Source/yii
```diff
@@ -8,8 +8,8 @@
  * @license http://www.yiiframework.com/license/
  */
 
-define('YII_DEBUG', false);
-define('YII_ENV', 'production');
+const YII_DEBUG = false;
```
mtompset, line 11: The define function is better, because if it is ALREADY DEFINED, it doesn't change it.

modules/silauth/src/Auth/Source/yii
```diff
@@ -8,8 +8,8 @@
...
+const YII_ENV = 'production';
```
mtompset, line 12: The define function is better, because if it is already defined, it doesn't change it.

modules/sildisco/public/metadata.php
```diff
@@ -57,7 +56,7 @@
     $keys[] = array(
         'type' => 'X509Certificate',
         'signing' => true,
-        'encryption' => ($hasNewCert ? false : true),
+        'encryption' => !$hasNewCert,
```
mtompset, line 59: Nice optimization.

modules/sildisco/public/metadata.php
```diff
@@ -220,14 +219,12 @@
...
         echo $metaxml;
-        exit(0);
     } else {
-
         header('Content-Type: text/html; charset=utf-8');
 
         echo '<pre>' . print_r($metaflat, true) . '</pre>';
-        exit(0);
     }
+    exit(0);
```
mtompset, line 227: Nice move.

modules/sildisco/src/Auth/Process/AddIdp2NameId.php
```diff
@@ -32,8 +33,6 @@ class AddIdp2NameId extends ProcessingFilter
 
     const SP_NAMEID_ATTR = 'saml:sp:NameID'; // The key for the NameID
 
-    const VALUE_KEY = 'Value';  // The value key for the NamedID entry
```
mtompset, line 35: Wasn't used anywhere?

modules/sildisco/src/IdPDisco.php
```diff
@@ -68,6 +69,8 @@ private function getSPEntityIDAndReducedIdpList(): array
 
     /**
      * @inheritDoc
+     * @throws MetadataNotFound|ConfigurationError
+     * @throws \Exception
```
mtompset, line 73: Is it a different Exception? Why the slash?

modules/sildisco/src/IdPDisco.php
```diff
@@ -96,7 +99,7 @@ public function handleRequest(): void
         $metadata = MetaDataStorageHandler::getMetadataHandler();
         $sp = $metadata->getMetaData($spEntityId, 'saml20-sp-remote');
 
-        $t = new Template($this->config, 'selectidp-links', 'disco');
+        $t = new Template($this->config, 'selectidp-links');
```
mtompset, line 102: Poor disco. No longer wanted. :)

features/fakes/FakeIdBrokerClient.php
briskt: Not sure how I missed those, thanks.

features/fakes/FakeIdBrokerClient.php
briskt: Not thoroughly, but thanks to your reminder, I'll check more. This one's OK. The real IdBrokerClient is already strong typed and we haven't seen any errors from it on production.

modules/mfa/public/send-manager-mfa.php
briskt: Agreed. Thanks for the tip.

modules/mfa/src/Auth/Process/Mfa.php
briskt: It comes from an external API, so we don't have control of it in this repo. In fact, I think it comes from two different services, which may be why it's a mixed type. And it gets passed to another API, so there's not much point in trying to fix it, except on a system level. Oh, the nastiness of microservices.

modules/mfa/src/Auth/Process/Mfa.php
briskt: No, I think that's just the style the original author decided to use to return only the first missing parameter in the error message. Obviously the `else`s aren't needed, but it wasn't flagged by lint, so I won't mess with it here.

modules/mfa/src/Auth/Process/Mfa.php
briskt: Are you sure? I didn't find any reference that backs that up.

modules/profilereview/src/Auth/Process/ProfileReview.php
briskt: No, it is not. When I did the upgrade to SimpleSAMLphp version 2, I had to strong type everything that was inherited because it was done in the library. So the changes I'm making now are not on inherited methods. Great question.

modules/silauth/src/Auth/Source/auth/AuthError.php
briskt: Hmm, didn't think about that. PHP is tricky. Thanks for the heads-up.

modules/silauth/src/Auth/Source/time/UtcTime.php
briskt: Yes, it's set in the constructor.

modules/silauth/src/Auth/Source/time/UtcTime.php
briskt: Yes, this also is set in the constructor.

modules/silauth/src/Auth/Source/yii
briskt: Well, PhpStorm didn't say anything about that. Thanks. But I don't think that's an issue here, since `yii` is always the first bit to run and these are at the top of the file. [This SO answer](https://stackoverflow.com/a/3193704/1461333) makes it sound like `const` is generally preferred.

modules/silauth/src/Auth/Source/yii
briskt: See above.

modules/sildisco/public/metadata.php
```diff
@@ -144,7 +143,7 @@
 
     if ($idpmeta->hasValue('OrganizationName')) {
         $metaArray['OrganizationName'] = $idpmeta->getLocalizedString('OrganizationName');
-        $metaArray['OrganizationDisplayName'] = $idpmeta->getLocalizedString(
+        $metaArray['OrganizationDisplayName'] = $idpmeta->getOptionalLocalizedString(
             'OrganizationDisplayName',
             $metaArray['OrganizationName']
```
briskt, line 148: Just to leave a note here: this was actually a bug since the SSP 2.0 upgrade changed the definition of `getLocalizedString` by removing the 2nd parameter. I missed this one during the upgrade effort earlier this year. Editorial: things like this make me scared to use interpreted languages.

modules/sildisco/src/Auth/Process/AddIdp2NameId.php
briskt: No, but I realize that it could be used outside this repo. I'll put it back...reluctantly.

modules/sildisco/src/IdPDisco.php
briskt: I think it's because of this: `yii\db\Exception`

modules/sildisco/src/IdPDisco.php
briskt: So sad. Another missed piece from the SSP 2.0 upgrade. That argument was `$defaultDictionary`. We only have one dictionary.

modules/profilereview/src/Auth/Process/ProfileReview.php
briskt: Yes. On your suggestion, I checked over them once more.

briskt (commented):
Thanks for diving into a big review!

modules/mfa/src/Auth/Process/Mfa.php
jason-jackson: The interesting thing is that [SimpleSamlPHP uses both ways](https://github.com/simplesamlphp/simplesamlphp/blob/ced50409a35cf05870335e7fa5c25afceeb00b3c/src/SimpleSAML/IdP.php#L355), with and without a pointer. (Linked one file as an example, they do it more than just there.)

jason-jackson (approved):
In general it seems fine. It's odd that you use both arrays and pointers to arrays, but then again SSP does that too. Might be nice to standardize.

modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -159,7 +161,7 @@ protected function getAttributeAllValues(string $attributeName, array $state): ?
      * Get an ID Broker client.
      *
      * @param array $idBrokerConfig
-     * @return IdBrokerClient
+     * @return IdBrokerClient|FakeIdBrokerClient
```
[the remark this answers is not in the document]
briskt: Neither is an interface in the OOP sense (`abstract` in PHP terms). That would actually be a good improvement. What is it about the union that you don't like?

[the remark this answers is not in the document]
mtompset: The interface idea is better, because what if there was a third id broker client?

modules/mfa/src/Auth/Process/Mfa.php
mtompset: The problem is if mfaId is not empty, it doesn't check the rest.

modules/mfa/src/Auth/Process/Mfa.php
mtompset: As long as the function call is the same as the inherited one, that's fine.

modules/profilereview/src/Auth/Process/ProfileReview.php
mtompset: Then you should NOT do that.

mtompset: Only strong type if the inherited functions are strong typed.

modules/silauth/src/Auth/Source/yii
mtompset: If you have a docker compose environment variable set, this would override that value. Define would not.

briskt: These aren't environment variables, though.

PR #284, opened 2024-10-10
Release 10.1.2 -- null coalesce to prevent a type error

briskt:
[IDP-1249](https://itse.youtrack.cloud/issue/IDP-1249) Clicking "Yes" when prompted to reset password (by expirychecker) triggers an error

---

### Fixed
- Add a null coalesce to prevent a type error.


hobbitronics (approved):
Nice find!

PR #288, opened 2024-11-06
remove fill-viewport from templates

hobbitronics:
image


hobbitronics:
> What does fill-viewport do?

sets width: 100vw; height: 100vh; but we already have height and width set to 100% in mdl-layout.

PR #289, opened 2024-11-07
Release - Remove unused css and re-center content in body

hobbitronics:
Fixed - The content was at the top which was only noticeable on bigger screen, but this centers is after having removed fill-viewport

image

### DIFF
--- a/modules/material/public/styles.2.3.6.css
+++ b/modules/material/public/styles.2.3.6.css
@@ -1,3 +1,7 @@
+body {
+  display: grid;
+}
+
 p {
   max-width: 60ch;
 }
@@ -5,11 +9,6 @@ p {
 pre {
   white-space: pre-line;
   max-width: 100%;
-}
-
-.fill-viewport {
-  width: 100vw;
-  height: 100vh;
 }
 
 [layout-children=row] {


modules/material/public/styles.2.3.6.css
```diff
@@ -1,3 +1,7 @@
+body {
+  display: grid;
```
briskt, line 2: CSS is so mysterious. How this achieves the effect you describe is not easy to understand. I'm just editorializing, not asking for an explanation necessarily.

briskt:
> Fixed - The content was at the top which was only noticeable on bigger screen, but this centers is after having removed fill-viewport

Your screenshot appears to be from a bigger screen. So in a smaller screen it would be scrollable?

briskt (approved):
Thanks for fixing this!

hobbitronics:
> > Fixed - The content was at the top which was only noticeable on bigger screen, but this centers is after having removed fill-viewport
> 
> Your screenshot appears to be from a bigger screen. So in a smaller screen it would be scrollable?

Yes, in mobile it is scrollable, but this change doesn't affect that. I'm not even sure why fill-viewport was making the announcemount higher than the viewport. This is probably the same as if we made body a flex column, or maybe used margin, although I already tried margin: auto which didn't work for some other reason. Too many layers of onion to get to the bottom.

modules/material/public/styles.2.3.6.css
hobbitronics: Maybe we should let ai do our css coding more and just treat it like magic.

briskt: I've heard it said that AI shouldn't have to create readable code...probably machine language would be best. It only does that for us stupid humans.

PR #290, opened 2024-11-13
Wrap calls to ga check for analytics_tracking_id

jason-jackson:
### Changed
- Wrap calls to ga check for analytics_tracking_id

### DIFF
--- a/modules/material/themes/material/default/selectidp-links.twig
+++ b/modules/material/themes/material/default/selectidp-links.twig
@@ -15,11 +15,15 @@
 
       document.querySelector('form').appendChild(idpInput);
 
-      ga('send', 'event', 'hub', 'choice', 'IdP', id);
+      {% if analytics_tracking_id is not empty %}
+        ga('send', 'event', 'hub', 'choice', 'IdP', id);
+      {% endif %}
     }
 
     function clickedAnyway(idpName) {
-      ga('send', 'event', 'hub', 'choice-disabled', 'IdP', idpName);
+      {% if analytics_tracking_id is not empty %}
+        ga('send', 'event', 'hub', 'choice-disabled', 'IdP', idpName);
+      {% endif %}
     }
   </script>
 </head>
--- a/modules/material/themes/material/silauth/loginuserpass.twig
+++ b/modules/material/themes/material/silauth/loginuserpass.twig
@@ -28,7 +28,9 @@
         });
       }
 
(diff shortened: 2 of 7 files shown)


briskt (approved):
Great job hunting down all the uses of `ga`!

briskt:
I'm surprised ga is not defined since we should be providing the tracking ID in all IdPs. There may be something else wrong with the ga library.

briskt:
I saw that @forevermatt marked GA as deprecated in idp-id-broker. Maybe he or @devon-sil can comment about its use here.

forevermatt:
> I saw that @forevermatt marked GA as deprecated in idp-id-broker. Maybe he or @devon-sil can comment about its use here.

In ID Broker, I was deprecating the use of Google Analytics for reporting MFA metrics (server side, from a cron job). This all looks like client-side reporting, but I don't know if we still use that, either. 🤷

modules/material/themes/material/silauth/loginuserpass.twig
```diff
@@ -100,9 +102,11 @@
             </span>
           </p>
 
-          <script>
-            ga('send', 'event', 'error', '{{ errorcode|e('js') }}', 'message', '{{ message|e('js') }}');
-          </script>
+          {% if analytics_tracking_id is not empty %}
```
mtompset, line 105: I feel like @forevermatt wanting an abstraction to make the sense of these "is not empty" clearer, even though I know what it is.

modules/material/themes/material/default/selectidp-links.twig
```diff
@@ -15,11 +15,15 @@
 
       document.querySelector('form').appendChild(idpInput);
 
-      ga('send', 'event', 'hub', 'choice', 'IdP', id);
+      {% if analytics_tracking_id is not empty %}
+        ga('send', 'event', 'hub', 'choice', 'IdP', id);
```
hobbitronics, line 19: I think this should be 
      gtag('event', 'IdP', {
        'event_category': 'hub',
        'event_label': 'choice',
        'value': id
      });

hobbitronics (changes requested):
I am finding `ga` is undefined, but in header.twig gtag is defined and has a slightly different api which looks like:
      gtag('event', 'IdP', {
        'event_category': 'hub',
        'event_label': 'choice',
        'value': id
      });

PR #292, opened 2024-11-26
fix google analytics and don't use don't use inline onclick

hobbitronics:


briskt:
Did you intend to close this?

PR #294, opened 2024-11-27
fix google analytics and replace inline onclick

hobbitronics:


briskt:
Looks like you started the `feature/fix-google-analytics` branch from `main` but the target (`feature/handle-missing-ga`) is a feature branch that isn't current with `main`. Do you want to change the target branch of this PR, or update that branch?

modules/material/themes/material/default/selectidp-links.twig
```diff
@@ -14,12 +14,34 @@
...
+        });
+      });
+    });
 
     function clickedAnyway(idpName) {
-      ga('send', 'event', 'hub', 'choice-disabled', 'IdP', idpName);
+      {% if analytics_tracking_id is not empty %}
+        gtag('event', 'IdP', {
+          'event_category': 'hub',
+          'event_label': 'choice',
```
briskt, line 41: Should the event_label be "choice-disabled"?

modules/material/themes/material/mfa/prompt-for-mfa-webauthn.twig
```diff
@@ -8,64 +8,69 @@
...
+      function createHiddenInput(name) {
+        const input = document.createElement('input');
+
+        input.type = 'hidden';
+        input.name = name;
+
+        return input;
+      }
+      
+      document.querySelector('.mdl-button.mdl-color-text--red').addEventListener('click', verifyWebAuthn);
```
briskt, line 67: Would it be better to add an id instead of using the class?

briskt (approved):
looks good, just a couple things to consider

PR #298, opened 2024-12-02
Fix WebAuthn prompt

forevermatt:
[IDP-1330](https://itse.youtrack.cloud/issue/IDP-1330) Email issue from ITSE Support Account

---

### Fixed
- Avoid waiting for a 2nd on-load event before calling `verifyWebAuthn()`

### Changed (non-breaking)
- Revise "Try again" button's ID for clarity
- Use "Try again" button's ID during `offerRetry()`, too

### DIFF
--- a/modules/material/themes/material/mfa/prompt-for-mfa-webauthn.twig
+++ b/modules/material/themes/material/mfa/prompt-for-mfa-webauthn.twig
@@ -37,7 +37,7 @@
       }
 
       function offerRetry() {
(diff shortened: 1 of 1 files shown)


jason-jackson (approved):

briskt:
This is a good solution. I did notice, though, that there were two other places where the same construct was used. One in `review.twig` and the other in `new-backup-codes.twig`. I filed a new issue and [PR](https://github.com/silinternational/ssp-base/pull/300) for this.

PR #300, opened 2024-12-03
Fix load listener

briskt:
[IDP-1340](https://itse.youtrack.cloud/issue/IDP-1340) ssp-base body event listener not triggered

### Fixed
- Directly call the `disableUnsupportedFeatures` and `prettifyDates` functions in the `DOMContentLoaded` listener instead of using `addEventListener` on the `body`.
- Enable printable code download on Edge.


briskt:
> Looks good. Thanks for cleaning up my mess. Btw do you have MFA set up locally?

No worries. Yes, I think so. Not sure I ever got it working with a fully-local setup though

PR #301, opened 2024-12-07
fix maskEmail to handle subdomains correctly

briskt:


modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -912,14 +912,9 @@ public static function maskEmail(string $email): string
...
-
-        $newEmail .= substr($domainA, 0, 1);
-        $newEmail .= str_repeat('*', strlen($domainA) - 1);
-        $newEmail .= '.';
-
-        $newEmail .= substr($domainB, 0, 1);
-        $newEmail .= str_repeat('*', strlen($domainB) - 1);
+        $newEmail .= implode('.', array_map(function ($part) {
+            return substr($part, 0, 1) . str_repeat('*', max(strlen($part) - 1, 0));
+        }, explode('.', $domain)));
```
hobbitronics, line 917: a bit hard for me to read, but maybe that's just me php. Looks like implode is taking 3 arguments but I only see the glue and the array in docs. Maybe something like this would be clearer: 
```
        $domainParts = explode('.', $domain);
        $maskedDomain = '';
        
        foreach ($domainParts as $part) {
          $firstCharacter = substr($part, 0, 1);
          $maskedPart = $firstCharacter . str_repeat('*', max(strlen($part) - 1, 0));
          $maskedDomain .= $maskedPart . '.';
        }
        $maskedDomain = rtrim($maskedDomain, '.');
        $newEmail .= $maskedDomain;```

**Review by hobbitronics — APPROVED** [review]
Looks good, just a bit hard to parse for my weak php so offering an alternative if you like.

File: modules/mfa/src/Auth/Process/Mfa.php (hunk 1 above)
Comment (hobbitronics) [reply modules/mfa/src/Auth/Process/Mfa.php:917]: Ahh, I see now.

Comment (jason-jackson) [reply modules/mfa/src/Auth/Process/Mfa.php:917]: I'm a bit hesitant about this too, I think it makes the function more complex than it needs to be. Maybe something like this would be better?

```
    public static function maskEmail(string $email): string
    {
        $newEmail = '';
        $useRealChar = true;
        /*
         * Replace all characters with '*', except
         * the first one, the last one, underscores, dashes, periods and each
         * character that follows and underscore.
         */
        foreach (str_split($email) as $nextChar) {
            if ($useRealChar) {
                $newEmail .= $nextChar;
                $useRealChar = false;
            } else if ($nextChar === '_' || $nextChar === '.' || $nextChar === '@') { // TODO should this also include `-` for JAARS?
                $newEmail .= $nextChar;
                $useRealChar = true;
            } else {
                $newEmail .= '*';
            }
        }

        $at = strpos($email, '@');
        $newEmail[$last] = $email[$last];

        return $newEmail;
```

briskt: I think I'll implement all three, with a random number generator to decide which one to use on each invocation.

jason-jackson: Interesting find, but the regex doesn't seem to handle `_` very well. Without including underscore in regex, it will mask it (`f*********@e******.c**`), and if I do include it, it won't unmask the following letter (`f****_****@e******.c**`).

modules/mfa/tests/MfaTest.php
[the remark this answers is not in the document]
briskt: Even though it's not behat? :)

modules/mfa/src/Auth/Process/Mfa.php
briskt: Thanks all. One day we may use AI-generated code when the machines have stripped away all of our responsibilities. Until that day, readable code is still a good thing.

PR #302, opened 2024-12-13
add remember_me_preference cookie

hobbitronics:
[IDP-105](https://itse.youtrack.cloud/issue/IDP-105) Set cookie to not check remember me box if user unchecks it when logging in so we don't default it checked again in the future.

### DIFF
--- a/modules/mfa/src/Auth/Process/Mfa.php
+++ b/modules/mfa/src/Auth/Process/Mfa.php
@@ -495,6 +495,8 @@ class Mfa extends ProcessingFilter
         // Set remember me cookies if requested
         if ($rememberMe) {
             self::setRememberMeCookies($state['employeeId'], $state['mfaOptions']);
+        } else {
+            self::clearRememberMeCookies();
         }
 
         $logger->warning(json_encode([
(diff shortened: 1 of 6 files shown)


modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -808,6 +810,33 @@ public static function setRememberMeCookies(
         setcookie('c2', $expireDate, $expireDate, '/', null, $secureCookie, true);
     }
 
+    /**
+     * Clear remember_me cookies (c1 and c2)
+     */
+    public static function clearRememberMeCookies(): void
+    {
+        $secureCookie = Env::get('SECURE_COOKIE', true);
+        setcookie('c1', '', time() - 3600, '/', null, $secureCookie, true);
```
briskt, line 819: just curious why you used `time() - 3600` and not just `time()`

briskt (approved):
good implementation

modules/mfa/src/Auth/Process/Mfa.php
hobbitronics: I guess it's considered best practice by some on [stacko](https://stackoverflow.com/questions/56722961/is-setting-cookie-expiration-time-time-3600-always-correct)

hobbitronics: It's even in the [docs](https://www.php.net/manual/en/function.setcookie.php)

briskt: Kinda seems like a "we've always done it that way" sort of thing. I can't figure out how it helps. If it's a time zone issue, shouldn't it be `time() - 3600 * 24`?

hobbitronics: You're probably right, but time drift of the client is also mentioned. Not sure how common that is these days.

PR #303, opened 2025-01-31
add desiredRememberMeState to local storage

hobbitronics:

- Fixes the checkbox becoming uncheck when switching between mfa types due to page reload

### DIFF
--- a/modules/material/themes/material/mfa/prompt-for-mfa-webauthn.twig
+++ b/modules/material/themes/material/mfa/prompt-for-mfa-webauthn.twig
@@ -9,19 +9,6 @@
 
   <script>
     document.addEventListener('DOMContentLoaded', function() {
-      const rememberMeCheckbox = document.getElementById('rememberMe');
-      
-      if (rememberMeCheckbox) {
-        const desiredRememberMeState = localStorage.getItem('desiredRememberMeState');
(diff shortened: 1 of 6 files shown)


modules/material/themes/material/mfa/prompt-for-mfa-webauthn.twig
```diff
@@ -9,6 +9,19 @@
...
+        const desiredRememberMeState = localStorage.getItem('desiredRememberMeState');
+        if (desiredRememberMeState === 'true') {
+          rememberMeCheckbox.checked = true;
+        }
+
+        rememberMeCheckbox.addEventListener('change', function() {
+          localStorage.setItem('desiredRememberMeState', rememberMeCheckbox.checked);
+        });
+      }
+
```
briskt, line 24: I feel like 4x is too much to duplicate this code. Could you drop a new js file in the modules/mfa/public folder and reference it like `<script src="{{ remember_me_js_path }}"></script>`, similar to the browser.js file?

hobbitronics: Yeah, not sure why I didn't do that to start with.

briskt (approved):

PR #304, opened 2025-01-31
Clear remember me from storage on submit and fix script path

hobbitronics:

### DIFF
--- a/modules/mfa/public/remember-me.js
+++ b/modules/mfa/public/remember-me.js
@@ -11,4 +11,11 @@ document.addEventListener('DOMContentLoaded', function() {
       localStorage.setItem('desiredRememberMeState', rememberMeCheckbox.checked);
     });
   }
+
+  const form = document.querySelector('form');
+  if (form) {
+    form.addEventListener('submit', function() {
+      localStorage.removeItem('desiredRememberMeState');
+    });
+  }
 });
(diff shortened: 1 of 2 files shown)


briskt (approved):

modules/mfa/public/remember-me.js
```diff
@@ -11,4 +11,11 @@ document.addEventListener('DOMContentLoaded', function() {
       localStorage.setItem('desiredRememberMeState', rememberMeCheckbox.checked);
     });
   }
+
+  const form = document.querySelector('form');
+  if (form) {
+    form.addEventListener('submit', function() {
+      localStorage.removeItem('desiredRememberMeState');
```
forevermatt, line 18: What's the reasoning behind clearing it? (I ask because I thought part of the goal was for it to stay how the user set it, even on subsequent logins, but that may have been in a separate YT issue.)

hobbitronics: So it will rely on the server side `rememberMePreference` cookie which I put in place earlier. The only issue is I think that will expire in 30 days possibly at the same time the rememberMe cookie expires so I could either remove the server side cookie and not remove this or I could set the expiration date later or indefinite. Any preference?

forevermatt: I don't see any need for this to be done server-side (since I am not aware of us storing/using that preference anywhere server side), so I would lean towards just using the client-side local-storage option, with no expiration. I think that fits our goal better.

PR #305, opened 2025-02-04
add ok button to start webauthn verification

hobbitronics:
[IDP-1401](https://itse.youtrack.cloud/issue/IDP-1401) Update 2SV webauthn flow to require user to click button to perform the action

## Added

- add a button to start webauthn verification instead of automoatically starting

### DIFF
--- a/modules/material/locales/en/LC_MESSAGES/material.po
+++ b/modules/material/locales/en/LC_MESSAGES/material.po
@@ -137,6 +137,9 @@ msgstr "USB key icon"
 msgid "{mfa:webauthn_instructions}"
 msgstr "You may now insert your security key and press its button."
 
+msgid "{mfa:webauthn_start}"
+msgstr "Click OK when you're ready to start."
+
 msgid "{mfa:webauthn_error_unknown}"
 msgstr "Something went wrong with that request, unable to verify at this time."
 
--- a/modules/material/locales/en/LC_MESSAGES/material.po
+++ b/modules/material/locales/en/LC_MESSAGES/material.po
@@ -135,7 +135,7 @@ msgid "{mfa:webauthn_icon}"
 msgstr "USB key icon"
 
 msgid "{mfa:webauthn_instructions}"
-msgstr "You may now insert your security key and press its button."
+msgstr "Follow the prompts to use your key."
 
 msgid "{mfa:webauthn_start}"
 msgstr "Click OK when you're ready to start."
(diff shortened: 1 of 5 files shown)


modules/material/locales/en/LC_MESSAGES/material.po
```diff
@@ -137,6 +137,9 @@ msgstr "USB key icon"
 msgid "{mfa:webauthn_instructions}"
 msgstr "You may now insert your security key and press its button."
```
briskt, line 138: I think this text may be unnecessary at best, and incorrect at worst. There are other types of webauthn keys besides a "security key" that should be inserted and touched. The browser does a better job at guiding the user. @forevermatt what do you think?

modules/material/themes/material/mfa/prompt-for-mfa-webauthn.twig
```diff
@@ -10,6 +10,7 @@
   <script>
     document.addEventListener('DOMContentLoaded', function() {
       function verifyWebAuthn() {
+        document.getElementById('instructions').textContent = '{{ '{mfa:webauthn_instructions}'|trans }}';
```
briskt, line 13: If we decide the instructions are unnecessary, this line could be removed.

modules/material/locales/en/LC_MESSAGES/material.po
forevermatt: We certainly need to update the wording, but I don't know what to change it to, nor whether now is the time to do that. Feel free to use your judgment on this one, based on how it looks in an actual browser.

modules/material/themes/material/mfa/prompt-for-mfa-webauthn.twig
```diff
@@ -37,9 +38,11 @@
       }
 
       function offerRetry() {
-        const retryButton = document.getElementById('retryButton');
+        const verifyBtn = document.getElementById('verifyBtn');
 
-        retryButton.classList.remove('hide');
+        verifyBtn.classList.add('mdl-color-text--red');
+        verifyBtn.classList.remove('mdl-color-text--primary');
+        verifyBtn.textContent = '{{ '{mfa:button_try_again}'|trans }}';
```
forevermatt, line 45: Do we need to do any explicit escaping in places like this where we're injecting twig values into JavaScript? I don't remember if Twig does that automatically, or if we need to add some other `|js` or something (that's a total guess).

hobbitronics, line 45:
```suggestion
        verifyBtn.textContent = '{{ '{mfa:button_try_again}'|trans|escape('js') }}';

```

modules/material/locales/en/LC_MESSAGES/material.po
hobbitronics: Since there is a delay as the browser prompts for the key it might be nice to have something. Maybe "You may now use your passkey"

briskt (approved):

PR #306, opened 2025-02-05
move options to card actions so menu doesn't overflow and use mdl-but…

hobbitronics:
…ton--raised

image

image


briskt (approved):
Looks great!

PR #309, opened 2025-02-19
Release 10.2.0 -- add Parameter Store as an alternative to AppConfig

briskt:
### Added
- Added Parameter Store as an alternative to AppConfig

### DIFF
--- a/dockerbuild/run.sh
+++ b/dockerbuild/run.sh
@@ -20,14 +20,12 @@ cd /data
 cat /etc/*release | grep PRETTY
 php -v | head -n 1
 
-if [[ -z "${APP_ID}" ]]; then
-  apache2ctl -k start -D FOREGROUND
+if [[ $PARAMETER_STORE_PATH ]]; then
+  config-shim --path $PARAMETER_STORE_PATH apache2ctl -k start -D FOREGROUND
+elif [[ $APP_ID ]]; then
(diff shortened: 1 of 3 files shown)


dockerbuild/run.sh
```diff
@@ -20,14 +20,12 @@ cd /data
...
+  config-shim --path $PARAMETER_STORE_PATH apache2ctl -k start -D FOREGROUND
+elif [[ $APP_ID ]]; then
+  config-shim --app $APP_ID --config $CONFIG_ID --env $ENV_ID apache2ctl -k start -D FOREGROUND
 else
-  if [[ -z "${STRATEGY_ID}" ]]; then
-    config-shim --app $APP_ID --config $CONFIG_ID --env $ENV_ID apache2ctl -k start -D FOREGROUND
-  else 
-    config-shim -u --strategy $STRATEGY_ID --app $APP_ID --config $CONFIG_ID --env $ENV_ID apache2ctl -k start -D FOREGROUND
-  fi
+  apache2ctl -k start -D FOREGROUND
```
hobbitronics, line 28: This might be a "dumb" question, but is --strategy never going to be used?

hobbitronics (approved):
Looks good, just a question of removing --strategy here.

dockerbuild/run.sh
briskt: Good eye. It was kind of an experiment that wasn't as helpful as I'd hoped.

PR #310, opened 2025-02-20
Release 10.2.1 -- publish a major-version tag also

briskt:


mtompset (approved):
Simple enough change.

PR #320, opened 2025-03-17
Add "send MFA recovery code" page

forevermatt:

---

Note:
- This PR gets the page to exist and have correct content, but the code to handle page submissions has not yet been added.
- Use of this is behind a "feature flag" of sorts (the presence of new environment variables), so it being incomplete at this point is not a problem.

---

### Added
- Add a local mock-API (for testing) using Mockoon 
- Add simple "API client" for calling recovery-contacts API
- Define test scenario for seeing manager and another MFA recovery contact

### Changed (non-breaking)
- Extract method for getting the (not masked) manager email address
  * I ended up not needing the unmasked manager email address yet, but decided not to go back through and undo that change.


modules/material/locales/en/LC_MESSAGES/material.po
```diff
@@ -164,6 +164,27 @@ msgstr "A temporary code was sent your recovery contact at %managerEmail%."
...
+msgstr "Please select one of the following:"
+
+msgid "{mfa:recovery_your_supervisor}"
+msgstr "your supervisor"
+
+msgid "{mfa:recovery_sent}"
+msgstr "A temporary code was sent to your recovery contact."
+
+msgid "{mfa:recovery_input}"
+msgstr "Enter code"
```
briskt, line 186: These two aren't referenced. Are they still needed?

modules/mfa/public/send-recovery-mfa.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
briskt, line 1: If you'd like to use the object-based controller mechanism that's new to SimpleSAMLphp 2.0, you can use the sildisco module as an example (or https://github.com/simplesamlphp/simplesamlphp/tree/master/modules/saml for a larger example). Actually, I'm not sure if it works to only add new actions to the class while leaving the existing ones in ordinary php files. But I thought I'd at least mention it in case you're interested.

I never found actual documentation on it. Maybe Symfony docs would be applicable. But the ultra-short TLDR is to create a route file in module/mfa/routing/routes/routes.yml and specify class methods in a class defined in module/mfa/src/Controller/Recovery.php

modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -289,6 +291,33 @@ public static function getNumBackupCodesUserHad(array $mfaOptions): int
...
+        $recoveryConfig = $state['recoveryConfig'] ?? [];
+        $emailAddressValues = $state['Attributes']['mail'] ?? [''];
+        $emailAddress = $emailAddressValues[0] ?? '';
+
+        $apiUrl = $recoveryConfig['api'] ?? '';
+        $apiKey = $recoveryConfig['apiKey'] ?? '';
+        $fallbackEmail = $recoveryConfig['fallbackEmail'] ?? '';
+        $fallbackName = $recoveryConfig['fallbackName'] ?? '';
+
+        $apiClient = new ApiClient($apiKey);
```
briskt, line 308: Would it be helpful to describe what API this is calling? Either a more specific variable name or a comment or both? I _think_ I know, but I'm not sure. Since this is a public repo and the other is not, maybe a name like `recoveryApiClient` is enough. The README has a bit of documentation on it, though it doesn't describe what the request or response should contain.

briskt (approved):
Looks good! I added some comments for things to consider if you are interested.

modules/material/locales/en/LC_MESSAGES/material.po
forevermatt: I expect the first to be used, so I'll leave it. The 2nd might not... good catch.

modules/mfa/public/send-recovery-mfa.php
forevermatt: Thanks! I don't think I will switch at this point, but I'm glad to know about that.

modules/mfa/src/Auth/Process/Mfa.php
forevermatt: Thanks for sharing that. I'll rename some variables here, to hopefully make it clearer.

As for the readme... I'm a bit confused, because I did (attempt to) specify there how the request works and what structure the response needs to have ,here:  
<https://github.com/silinternational/ssp-base?tab=readme-ov-file#new-recovery-contacts-feature-coming-soon>

Please feel free to share any additional details that you would find helpful to have in the readme file about this.

forevermatt: @briskt I'm going to go ahead and resolve this for now (so I can proceed with the merge, since you approved the PR). Feel free to continue this conversation, though, and I can make additional requested changes in future PRs.

briskt: I'm sorry. I didn't read the README carefully enough. You had good documentation already.

PR #323, opened 2025-03-19
Send recovery contact email address to ID Broker

forevermatt:
[IDP-1424](https://itse.youtrack.cloud/issue/IDP-1424) Add a "send to my mail admin" option to ssp-base's "mfa" module

---

### WAITING FOR
- #322 

---

### Changed (non-breaking)
- Have the FakeIdBrokerClient `mfaCreate()` require a recovery email for `recovery` MFAs
  * This is because we expect ID Broker to allow such behavior soon. See [IDP-1482](https://itse.youtrack.cloud/issue/IDP-1482).
- Structure the recovery contact data used in the view as name => email

### Fixed
- Pass the selected recovery contact's email to ID Broker's `mfaCreate()`

PR #322, opened 2025-03-19
Requesting MFA code be sent to a recovery contact

forevermatt:
### Added
- Add test for submitting request to send code to a recovery contact
- Adapt `sendManagerCode()` to create a similar `sendRecoveryCode()`
- Adapt `prompt-for-mfa-manager` view template to create a `prompt-for-mfa-recovery` view template

### Fixed
- Ensure local IDP4 has the FakeIdBroker Client, for testing
- Update FakeIdBrokerClient to handle `recovery` MFA type
- Re-add `mfa:recovery_input` translation text

### Changed (non-breaking)
- If the user has a `recovery` MFA option, default to asking for that (even before a `manager` MFA option)

---

### Note
- The `Mfa::sendRecoveryCode()` function does not yet use the `$mfaRecoveryContactID` passed to it. That will be addressed in future updates to the tests.

PR #324, opened 2025-03-19
Abbreviate recovery contacts' names in the view

forevermatt:

---

### WAITING FOR
- #323
  * ... or maybe just #322

---

### Added
- Enable the test code to verify what the recovery contacts API returns

### Fixed
- Abbreviate recovery contacts' names in the form shown to the user
- Confirm that recovery contacts' names are shown abbreviated

### DIFF
--- a/modules/mfa/src/Auth/Process/Mfa.php
+++ b/modules/mfa/src/Auth/Process/Mfa.php
@@ -128,10 +128,13 @@ class Mfa extends ProcessingFilter
         }
     }
 
-    private static function abbreviateName(string $name): string
+    public static function abbreviateName(string $name): string
     {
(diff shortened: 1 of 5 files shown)


modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -128,10 +128,13 @@ public static function validateConfigValue(string $attribute, mixed $value, Logg
         }
     }
 
-    private static function abbreviateName(string $name): string
+    public static function abbreviateName(string $name): string
```
briskt, line 131: I function header would be helpful, I think.

briskt (approved):

PR #327, opened 2025-03-24
Combine MFA recovery test scenarios for sending code to manager and non-manager

forevermatt:
### WAITING FOR
- #326

---

### Fixed
- Combine MFA recovery tests for sending code to manager and non-manager

PR #328, opened 2025-03-25
Fallback recovery contact

forevermatt:

---

### WAITING FOR
- #327

---

### Fixed
- Add test ensuring we use fallback option if API returns no recovery contacts
- Stop abbreviating the recovery-contacts fallback name

### DIFF
--- a/README.md
+++ b/README.md
@@ -487,7 +487,8 @@ and `email` field. Example:
 
 Names returned by the API will be partially abbreviated to avoid giving out too
 much information in case a user's password is compromised. If the returned array
-is empty, the provided fallback parameters will be used.
+is empty, the provided fallback parameters will be used. Note: The fallback name
+will not be abbreviated.
 
 #### Why use an AuthProc for MFA?
 Based on...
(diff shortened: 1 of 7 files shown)


mtompset (approved):

README.md
```diff
@@ -487,7 +487,8 @@ and `email` field. Example:
 
 Names returned by the API will be partially abbreviated to avoid giving out too
 much information in case a user's password is compromised. If the returned array
-is empty, the provided fallback parameters will be used.
+is empty, the provided fallback parameters will be used. Note: The fallback name
+will not be abbreviated.
```
briskt, line 491: I assume that this was a decision that was discussed and agreed upon. To me it seems like a surprising inconsistency. I'm not saying the decision must be reversed, but I get a bit uncomfortable with unexpected complexity.
Is this detail covered in tests? For instance, a test that ensures that the name is abbreviated if it comes from the API, and another that ensures it is NOT abbreviated when the fallback is used?

briskt (approved):

README.md
forevermatt: It was not discussed, so thanks for communicating your concerns.

I realized that abbreviating the fallback name (which is the same for everyone in the IDP, and is set via an environment variable) makes it very hard for us to control exactly what is shown in the UI for the fallback contact. For example, "Google Admins" or "IT Help Desk": We couldn't show either of those as the fallback name if we were abbreviating the fallback name.

We abbreviate the names returned by the API because those are probably actual people's names, and we don't know ahead of time exactly what they will be. The fallback name is exactly whatever we set it to be, so forcing it to be abbreviated seemed unhelpful.

PR #329, opened 2025-03-26
Add test for when the recovery-contacts API does return a contact

forevermatt:

---

### WAITING FOR
- #328

---

### Fixed
- Add test for when the recovery-contacts API does return a contact

### DIFF
--- a/development/idp4-local/config/authsources.php
+++ b/development/idp4-local/config/authsources.php
@@ -120,6 +120,43 @@ $config = [
                 ],
                 'manager_email' => ['manager@example.com'],
             ],
+
+            /**
+             * A user with MFA configured (in this case, backup codes) a
+             * manager, and a recovery contact (in the recovery-contacts mock
+             * API).
+             */
+            'has_backupcode_mgr_recovery_contact:a' => [
+                'eduPersonPrincipalName' => ['has_backupcode_mgr_recovery_contact@mfaidp'],
+                'eduPersonTargetID' => ['4927f678-6d1c-423f-a094-46c6b56f7bc5'], // Make it unique
+                'sn' => ['Recovery Contact'],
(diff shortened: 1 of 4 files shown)


development/idp4-local/config/authsources.php
```diff
@@ -120,6 +120,43 @@
...
+             * A user with MFA configured (in this case, backup codes) a
+             * manager, and a recovery contact (in the recovery-contacts mock
+             * API).
+             */
+            'has_backupcode_mgr_recovery_contact:a' => [
+                'eduPersonPrincipalName' => ['has_backupcode_mgr_recovery_contact@mfaidp'],
+                'eduPersonTargetID' => ['4927f678-6d1c-423f-a094-46c6b56f7bc5'], // Make it unique
+                'sn' => ['Recovery Contact'],
+                'givenName' => ['Has Backupcode Mgr'],
+                'mail' => ['has_backupcode_mgr_recovery_contact@example.com'],
```
mtompset, line 134: This says 'mail', but the other code above says 'email'. Are they related?

briskt (approved):

development/idp4-local/config/authsources.php
forevermatt: Sorry, where do you see `email`?

The SAML attribute name is `mail`, so that is what these email addresses in the `authsources.php` files should have as their key.

forevermatt: Oh, are you referring to the mfa-recovery-contacts.json file? Yes, that one is supposed to be `email` (because that is the query string parameter for the recovery contacts API "spec" outlined in this code's readme file), and yes, this `authsources.php` file is supposed to use `mail` as the attribute name. Thanks for checking.

PR #330, opened 2025-03-26
Fix the `iClickTheRequestAssistanceLink()` test step

forevermatt:

---

### WAITING FOR
- #329

---

### Fixed
- Fix the `iClickTheRequestAssistanceLink()` test step

### DIFF
--- a/features/bootstrap/MfaContext.php
+++ b/features/bootstrap/MfaContext.php
@@ -595,9 +595,13 @@ class MfaContext extends FeatureContext
      */
     public function iClickTheRequestAssistanceLink()
     {
-        // find image of the recovery contact option presented in prompt_for_mfa_manager.php
-        $printableCodeOption = $this->session->getPage()->find('css', 'img[src=mfa-manager\002Esvg]');
-        $printableCodeOption->click();
+        $page = $this->session->getPage();
+        $helpOption = $page->findById('more-options-manager');
+        Assert::assertNotNull(
+            $helpOption,
+            'Could not find the "I need help" option'
+        );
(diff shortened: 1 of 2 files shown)


features/bootstrap/MfaContext.php
```diff
@@ -595,9 +595,13 @@ public function iShouldNotSeeALinkToSendACodeToTheUsersManager()
      */
     public function iClickTheRequestAssistanceLink()
     {
-        // find image of the recovery contact option presented in prompt_for_mfa_manager.php
-        $printableCodeOption = $this->session->getPage()->find('css', 'img[src=mfa-manager\002Esvg]');
-        $printableCodeOption->click();
+        $page = $this->session->getPage();
+        $helpOption = $page->findById('more-options-manager');
```
mtompset, line 599: 23 below does a concatenation, which may or may not be manager. Perhaps multiple findById's consolidated together before doing an assertNotNull?

briskt (approved):

features/bootstrap/MfaContext.php
forevermatt: Correct, but the "I need help" option is (currently) always given a `type` of "manager". If we change that in the future, we can fix this test if/when it fails (due to the absence of such an element on the page). I'd like to proceed with this code, unless you think this is a significant problem.

forevermatt: @mtompset I'm going to go ahead and resolve this conversation so I can proceed with the merge (since it's been approved), but please feel free to ask about this further if you have additional questions or would like more explanation.

PR #336, opened 2025-04-15
Release 10.2.17 - Log the MFA recovery options (when viewed)

forevermatt:
### Fixed
- Log what MFA recovery contacts were returned/offered and the (unmasked) manager email

### DIFF
--- a/modules/mfa/src/Auth/Process/Mfa.php
+++ b/modules/mfa/src/Auth/Process/Mfa.php
@@ -719,6 +719,7 @@ class Mfa extends ProcessingFilter
 
         $state['mfaOptions'] = $mfaOptions;
         $state['managerEmail'] = self::getMaskedManagerEmail($state);
+        $state['unmaskedManagerEmail'] = self::getManagerEmail($state);
         $state['idBrokerConfig'] = [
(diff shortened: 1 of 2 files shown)


modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -719,6 +719,7 @@ protected function redirectToMfaPrompt(array &$state, string $employeeId, array
 
         $state['mfaOptions'] = $mfaOptions;
         $state['managerEmail'] = self::getMaskedManagerEmail($state);
+        $state['unmaskedManagerEmail'] = self::getManagerEmail($state);
```
briskt, line 722: Do you know if the state can be decoded on the client side?

briskt (approved):

modules/mfa/src/Auth/Process/Mfa.php
forevermatt: I believe the state is exclusively server-side. Data from it is used to render the HTML returned to the client, but I think it's only stored in the session on the server, not sent to the client.

Inspecting the response from the server lines up with that understanding: none of the state data seems to have arrived at the client other than what was intentionally included in the HTML.

Thanks for checking.

PR #338, opened 2025-04-16
Log the (intended) recipient of MFA recovery code

forevermatt:
### Fixed
- Log who the MFA recovery code was sent to
- If we failed to send an MFA recovery code, log who it would have gone to


forevermatt:
For error situations, if you think "recipient" is a misleading title for who _would have_ received the recovery code, feel free to suggest something else.

mtompset (approved):

briskt (approved):
I think "recipient" is fine. In the error context, it's not too big of a leap to assume that the email address never received the message.

PR #339, opened 2025-04-16
Fix labels for "use recovery" vs "use manager"

forevermatt:
### Fixed
- Add missing `mfa:use_recovery` string and translations
- Modify the `mfa:use_manager` translations, to differ from `mfa:use_recovery`
    * I'm concerned that the Korean translation changed more than just the "recovery contact" to "supervisor" part, since the new text is so different. It seems to now have a different level of formality (e.g. it now seems to say the equivalent of "Please use..." instead of "Use..."). I don't know enough about Korean to verify or fix that, though.


mtompset (approved):

briskt (approved):
Would it be worth finding a native Korean speaker to help with this?

PR #340, opened 2025-05-01
Help user know which manager it is (in the recovery-contacts list)

forevermatt:
[IDP-1536](https://itse.youtrack.cloud/issue/IDP-1536) Include the masked manager email address after "your supervisor" when selecting a recovery contact

---

### Fixed
- Show the (masked) manager email to help identify which supervisor
- Rename view's variable to clarify that it is the masked manager email

### DIFF
--- a/modules/material/themes/material/mfa/send-recovery-mfa.twig
+++ b/modules/material/themes/material/mfa/send-recovery-mfa.twig
@@ -43,9 +43,9 @@
               <br />
             {% endfor %}
 
-            {% if manager_email is not empty %}
+            {% if masked_manager_email is not empty %}
               <input type="radio" name="mfaRecoveryContactID" id="option-manager" value="recovery-contact-id-manager">
-              {{ '{mfa:recovery_your_supervisor}'|trans }} ({{ manager_email }})
+              {{ '{mfa:recovery_your_supervisor}'|trans }} ({{ masked_manager_email }})
               <br />
             {% endif %}
           </p>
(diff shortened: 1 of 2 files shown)


modules/material/themes/material/mfa/send-recovery-mfa.twig
```diff
@@ -43,9 +43,9 @@
               <br />
             {% endfor %}
 
-            {% if manager_email is not empty %}
+            {% if masked_manager_email is not empty %}
```
mtompset, line 46: Why the variable rename?

modules/mfa/public/send-recovery-mfa.php
```diff
@@ -63,7 +63,7 @@
 
 $t = new Template($globalConfig, 'mfa:send-recovery-mfa');
 $t->data['recovery_contacts_by_name'] = $recoveryContactsForView;
-$t->data['manager_email'] = $state['managerEmail'];
+$t->data['masked_manager_email'] = $state['managerEmail'];
```
mtompset, line 66: Is that value masked in the first place?

mtompset (approved):
The rename reason is not obvious in this code, but adding the email address (masked or not) is a good idea.

modules/material/themes/material/mfa/send-recovery-mfa.twig
briskt: In the PR description:

> Rename view's variable to clarify that it is the masked manager email

modules/mfa/public/send-recovery-mfa.php
briskt: In Mfa.php:

```php
$state['managerEmail'] = self::getMaskedManagerEmail($state);
```

Maybe the state array key needs to be renamed also? (I'm fine if it's left as-is.)

briskt (approved):

modules/mfa/public/send-recovery-mfa.php
forevermatt: I would like to rename it in the state array as well, but that was a slightly more in-depth change than I wanted to do right now.

PR #344, opened 2025-05-09
Release 10.2.23 -- Update a Korean translation

briskt:
[IDP-1537](https://support.gtis.sil.org/issue/IDP-1537) Have somone verify the Korean on a new translated message

### Changed
- Updated the Korean translation for "Use code from my supervisor" based on advice from a Korean native.


hobbitronics (approved):
Looks good

PR #345, opened 2025-05-19
Adjust log levels for MFA-recovery option page loads

forevermatt:
### Fixed
- Adjust recovery-options-offered log message level to Notice 
  * "Notice" is for "Normal but significant events", and is included in our logs by default. "Info" level logs are not, and this seems like more than merely an info-level log event.
- Similarly log manager-mfa-code-offered message as Notice level 
  * This seems comparable to the recovery-options-offered message, so I am setting it to have the same log level.


mtompset (approved):
change from info to notice. Simple enough.

briskt (approved):

PR #347, opened 2025-07-01
upgrade php and dependencies

jason-jackson:
### Changed
- Added jq to Dockerfile for `make depsupdate` command to list deps
- Changed `str*` to `mb_str*` for usernames
- Upgraded php to 8.3
- Updated dependencies


briskt (approved):

mtompset (approved):
I can't remember all the things I looked for while upgrading MailAdmin, but mb_strtolower is a good start, if there is anything else left.

PR #348, opened 2025-07-17
Release 10.4.0 -- load the AWS RDS cert bundle for SSL DB connection

briskt:
### Added
- Added startup code to save a pem file from the base64-encoded environment variable created by Terraform.
- Added SSL configuration in all places the database is accessed.
- Added test key and cert for the test database to verify SSL connection is working.

### Changed
- Changed the ssp-idp1.local start command so that the cert load can happen before migration. It now uses the run-idp.sh script which runs migrations then runs run.sh. This required a few changes to the startup timing, which was better accomplished using Docker Compose health checks, and not our whenavail script.

_This is unnecessarily complex because of the multiple startup scripts ("hub" mode, IdP mode, and various tests) and multiple section of code that have their own database initialization. It would be great if we could simplify these things._


hobbitronics (approved):
Approved based on this being a Release

PR #351, opened 2025-09-16
Release 11.0.0 - Upgrade to SimpleSamlPHP 2.4.X

jason-jackson:
### Changed
- Upgraded to SimpleSamlPHP 2.4.X
- Switch from silinternational to sil-org
- Renamed docker-compose.yml to compose.yaml

This will be a major version release due to ACS and SLO being required to be an array. Strings are no longer allowed. See SSP release notes:
- https://simplesamlphp.org/docs/2.3/simplesamlphp-upgrade-notes-2.3.html (breaking changes)
- https://simplesamlphp.org/docs/2.4/simplesamlphp-upgrade-notes-2.4.html


installed-packages.json
```diff
@@ -117,171 +125,195 @@
...
     "version": "v1.4.0"
   },
+  {
+    "name": "simplesamlphp/composer-xmlprovider-installer",
+    "version": "v1.0.2"
+  },
   {
     "name": "simplesamlphp/saml2",
-    "version": "v4.18.1"
+    "version": "v5.0.4"
```
briskt, line 168: I wonder if they actually follow semantic versioning in simplesamlphp/saml2 since it's a developer-facing library unlike simplesamlphp/simplesamlphp.

dockerbuild/config/config.php
```diff
@@ -915,7 +915,7 @@
      */
     'language.available' => [
         'en', 'no', 'nn', 'se', 'da', 'de', 'sv', 'fi', 'es', 'ca', 'fr', 'it', 'nl', 'lb',
-        'cs', 'sk', 'sl', 'lt', 'hr', 'hu', 'pl', 'pt', 'pt-br', 'tr', 'ja', 'zh', 'zh-tw',
+        'cs', 'sk', 'sl', 'lt', 'hr', 'hu', 'pl', 'pt', 'pt_BR', 'tr', 'ja', 'zh', 'zh_TW',
```
briskt, line 918: Was this the only change in the config.php.dist template? I'm just wondering how you found this. If possible, do a diff against the [config.php.dist](https://github.com/simplesamlphp/simplesamlphp/blob/master/config/config.php.dist) file and update the comment at the top of this file. Also, please do the same for [UserPass.php](https://github.com/sil-org/ssp-base/blob/97e41537a546c5b36c9a59aa0ef648e962db81b7/development/UserPass.php#L5).

briskt (approved):
Good work, thanks for making these updates.

jason-jackson:
> > * Renamed docker-compose.yml to compose.yaml
> 
> This may be another band-aid to rip off. It would be best if we're consistent throughout all our repos.

`compose.yaml` is their recommended default, so I've been changing to that.

dockerbuild/config/config.php
jason-jackson: Updated `config.php`. There are no changes needed for `UserPass.php`, would you like me to comment on that anyway?

installed-packages.json
jason-jackson: ¯\_(ツ)_/¯

composer.json
```diff
@@ -14,14 +14,14 @@
     "ext-gmp": "*",
     "ext-json": "*",
     "codemix/yii2-streamlog": "^1.3",
-    "simplesamlphp/simplesamlphp": ">=2.2.2 <2.3.0",
+    "simplesamlphp/simplesamlphp": ">=2.2.2 <2.5.0",
```
forevermatt, line 17: Do you want to force it to be a 2.4 version (rather than still allowed 2.2 and 2.3)? I don't know if any of the changes we're making to be compatible with 2.4 are possibly not backwards compatible, but it seems like we wouldn't want to downgrade to a pre-2.4 version at this point.
```suggestion
    "simplesamlphp/simplesamlphp": ">=2.4.0 <2.5.0",
```

PR #395, opened 2026-04-02
Add tests for last_login_utc update scenarios (MFA authproc)

samuel-sowpati-sil:
[IDP-1807](https://support.gtis.sil.org/issue/IDP-1807) Consider adding tests to ssp-base for when to update last_login_utc

---

### Added

- `Mfa::redirectToMfaSetup()` now calls updateUserLastLogin() before redirecting (implements scenario 5).
PHPUnit unit tests in modules/mfa/tests/MfaTest.php covering [IDP-1807](https://support.gtis.sil.org/issue/IDP-1807)  scenarios 2 and 5.

### Note
- This PR addresses the `MFA Authproc` alone. Will work on testing scenarios for other modules like `silauth` in a different PR.

---


mtompset:
The tests failed, @samuel-sowpati-sil . I am re-running them to see if it was a temporary glitch, or if it is something that you need to fix.

samuel-sowpati-sil:
> The tests failed, @samuel-sowpati-sil . I am re-running them to see if it was a temporary glitch, or if it is something that you need to fix.

I'm trying to address the `Page not loaded (Behat\Mink\Exception\DriverException)` error that I initially received across different scenarios each time I run `make test`.

features/fakes/FakeIdBrokerClient.php
```diff
@@ -16,6 +16,8 @@ class FakeIdBrokerClient
 
     const RATE_LIMITED_MFA_ID = 987;
 
+    const LAST_LOGIN_UPDATED_FILE = '/data/features/fakes/last-login-updated.txt';
```
jason-jackson: Why not just store them as an array here instead of in a file?

dockerbuild/run-integration-tests.sh
```diff
@@ -12,4 +12,6 @@ whenavail "ssp-hub.local"  80 15 echo Hub ready
 whenavail "ssp-idp1.local" 80 20 echo IDP 1 ready
 whenavail "ssp-sp1.local"  80 5 echo SP 1 ready
 
-behat --no-interaction
```
jason-jackson, line 15: What are the reasons for making this change? It's helpful to see all failures rather than stopping at the first one.

jason-jackson (approved):
Tests look fine, some could be combined, but not really much point right now with only two.

I'd rather not:
- Add the `--stop-on-failure`
- Use a file for tests, consider tracking it via the FakeIdBrokerClient instead

Once those are reverted/fixed, I think it should be fine.

dockerbuild/run-integration-tests.sh
samuel-sowpati-sil: I initially added `--stop-on-failure` in commit thinking fail-fast would help debugging. But you're right—seeing all failures at once is essential for understanding scope

features/fakes/FakeIdBrokerClient.php
samuel-sowpati-sil: The `FakeIdBrokerClient.updateUserLastLogin()` runs in the ssp-idp1.local container (during auth), while test assertions run in the test container. These are separate processes, and the static arrays don't share state across containers.

The file works because both containers mount `./features:/data/features` (compose.yaml lines 71, 179), making the filesystem the communication bridge between them. Without shared storage, tests can't verify what happened during authentication.

samuel-sowpati-sil: Most tests in this codebase verify UI-observable outcomes (page content), but `updateUserLastLogin` is a silent background call with no UI feedback. Happy to discuss other approaches if you see something I missed :smiley:!

dockerbuild/run-integration-tests.sh
briskt: I just recently changed this from what you have, because the behat.yml file now has the stop-on-failure configured as the default. So you would need to remove it there also if we actually do NOT want to stop on failure. For what it's worth, I usually like to see only one failure. Many times it is the same root cause and fixing one fixes them all.

features/fakes/FakeIdBrokerClient.php
briskt: You've taken on quite a challenge with this task, and the creativity to use a file is impressive. I agree with @jason-jackson that using a file is not desirable, maybe because it is a rare, last-resort kind of hack. It adds an unwanted file on the host file system, indicated by the need to add to the .gitignore file.

Since this is really an integration test, the best solution is to use a full representation of the real system, including a real broker and not a mock. You then could send an http request to the broker asking for the real last-login time. But I am not sure how much additional work that would require, and whether it's worth the effort. Honestly, I am not sure this feature is important enough to deserve an integration test if we have unit test coverage.

briskt:
> ## [IDP-1807](https://support.gtis.sil.org/issue/IDP-1807) Consider adding tests to ssp-base for when to update last_login_utc
> ### Note
> * This PR addresses the `MFA Authproc` alone. Will work on testing scenarios for other modules like `silauth` in a different PR.
> 
> ### Added
> * Behat feature file (`features/last-login.feature`) with 4 scenarios covering
>   when `last_login_utc` should and should not be updated:
>   
>   * No 2SV required → last login updated
>   * 2SV required and passed → last login updated
>   * 2SV required but failed → last login NOT updated
>   * 2SV required but not yet set up, user sent to set up 2SV → last login updated
> * `LastLoginContext.php` Behat context with step definitions and a  `@BeforeScenario` hook to reset tracking state between scenarios
> * `last_login_features` suite entry in `behat.yml`
> 
> ### Changed (non-breaking)
> * `FakeIdBrokerClient`: `updateUserLastLogin()` now writes the employee ID to a
>   temp file so tests can assert whether it was called.
> * added  `getUpdatedLastLoginEmployeeIds()` and `clearLastLoginUpdatedFile()` static helpers
> * `.gitignore`: added `features/fakes/last-login-updated.txt` to exclude the  runtime tracking file generated by the fake client during test runs
> 
> ### Fixed
> * `dockerbuild/run-integration-tests.sh`: restored `./vendor/bin/behat` path  and `--stop-on-failure` flag, which were removed in commit `3831c22` causing  `make test` to fail with `behat: command not found`

I find it harder to read a PR description that assumes the headers are the beginning of a sentence completed by the bullet points. Since it doesn't always work that way, (such as the bullets in your Changed section), it is better to consistently write full sentences in each bullet. Less detail is acceptable also, if it helps to compensate for the additional work of writing full sentences. Thank you for writing good PR summaries and not cutting corners!

briskt:
> * `dockerbuild/run-integration-tests.sh`: restored `./vendor/bin/behat` path  and `--stop-on-failure` flag, which were removed in commit `3831c22` causing  `make test` to fail with `behat: command not found`

Did you rebuild your Docker container? The Dockerfile was changed to add /data/vendor/bin to the PATH.

modules/mfa/src/Auth/Process/Mfa.php
```diff
@@ -628,6 +628,17 @@ public static function redirectToMfaSetup(array $state): void
             var_export($mfaSetupUrl, true)
         ));
 
+        try {
+            $idBrokerClient = self::getIdBrokerClient($state['idBrokerConfig']);
+            $idBrokerClient->updateUserLastLogin($state['employeeId']);
```
briskt, line 633: Why is this being added? Did the tests reveal a bug? Seems that way, but you didn't mention it in the PR description.

briskt:
> The tests failed, @samuel-sowpati-sil . I am re-running them to see if it was a temporary glitch, or if it is something that you need to fix.

It may be the instability I've been chasing. ([IDP-1925](https://support.gtis.sil.org/issue/IDP-1925) ssp-base test failure: Don't prompt for MFA)

briskt (approved):
Approved, nice work.

I think you can restore the run-integration-tests.sh to the simpler `behat --no-interaction`. Just rebuild your local image.

features/fakes/FakeIdBrokerClient.php
samuel-sowpati-sil: > You've taken on quite a challenge with this task, and the creativity to use a file is impressive. I agree with @jason-jackson that using a file is not desirable, maybe because it is a rare, last-resort kind of hack. It adds an unwanted file on the host file system, indicated by the need to add to the .gitignore file.
> 
> Since this is really an integration test, the best solution is to use a full representation of the real system, including a real broker and not a mock. You then could send an http request to the broker asking for the real last-login time. But I am not sure how much additional work that would require, and whether it's worth the effort. Honestly, I am not sure this feature is important enough to deserve an integration test if we have unit test coverage.

You're right Steve. Taking your's and Jason's suggestion, I dropped the Behat test entirely and replaced it with PHPUnit unit tests in `modules/mfa/tests/MfaTest.php` PHPUnit runs in a single process, so a `SpyIdBrokerClient` with a static array works cleanly. The spy is injected via the existing `idBrokerClientClass` config option — no production code changed. All the Behat scaffolding and file-tracking additions are reverted.

samuel-sowpati-sil:
> > ## [IDP-1807](https://support.gtis.sil.org/issue/IDP-1807) Consider adding tests to ssp-base for when to update last_login_utc
> > ### Note
> > 
> > * This PR addresses the `MFA Authproc` alone. Will work on testing scenarios for other modules like `silauth` in a different PR.
> > 
> > ### Added
> > 
> > * Behat feature file (`features/last-login.feature`) with 4 scenarios covering
> >   when `last_login_utc` should and should not be updated:
> >   
> >   * No 2SV required → last login updated
> >   * 2SV required and passed → last login updated
> >   * 2SV required but failed → last login NOT updated
> >   * 2SV required but not yet set up, user sent to set up 2SV → last login updated
> > * `LastLoginContext.php` Behat context with step definitions and a  `@BeforeScenario` hook to reset tracking state between scenarios
> > * `last_login_features` suite entry in `behat.yml`
> > 
> > ### Changed (non-breaking)
> > 
> > * `FakeIdBrokerClient`: `updateUserLastLogin()` now writes the employee ID to a
> >   temp file so tests can assert whether it was called.
> > * added  `getUpdatedLastLoginEmployeeIds()` and `clearLastLoginUpdatedFile()` static helpers
> > * `.gitignore`: added `features/fakes/last-login-updated.txt` to exclude the  runtime tracking file generated by the fake client during test runs
> > 
> > ### Fixed
> > 
> > * `dockerbuild/run-integration-tests.sh`: restored `./vendor/bin/behat` path  and `--stop-on-failure` flag, which were removed in commit `3831c22` causing  `make test` to fail with `behat: command not found`
> 
> I find it harder to read a PR description that assumes the headers are the beginning of a sentence completed by the bullet points. Since it doesn't always work that way, (such as the bullets in your Changed section), it is better to consistently write full sentences in each bullet. Less detail is acceptable also, if it helps to compensate for the additional work of writing full sentences. Thank you for writing good PR summaries and not cutting corners!

My apologies, I'll keep that in mind.

samuel-sowpati-sil:
> > * `dockerbuild/run-integration-tests.sh`: restored `./vendor/bin/behat` path  and `--stop-on-failure` flag, which were removed in commit `3831c22` causing  `make test` to fail with `behat: command not found`
> 
> Did you rebuild your Docker container? The Dockerfile was changed to add /data/vendor/bin to the PATH.

I hadn't — after make clean && make dev the bare `behat` command resolves correctly via the Dockerfile PATH. Also dropped --stop-on-failure since behat.yml already sets stop_on_failure: true as the default;

modules/mfa/src/Auth/Process/Mfa.php
samuel-sowpati-sil: Not a bug — this is in-scope for `IDP-1807`, scenario 5: "Correct username/password, required to go set up 2SV → update last-login." redirectToMfaSetup() was the only path missing the updateUserLastLogin() call. The try/catch ensures a broker failure can't block the user from reaching the setup page. Should have called this out in the original PR description .

modules/mfa/tests/MfaTest.php
```diff
@@ -22,4 +29,118 @@ public function testMaskEmail()
...
+}
+
+/**
+ * Spy client used by MfaTest. Records each updateUserLastLogin() call in a
+ * static array so the test can assert on it.
+ *
+ * Lives in this file because it has only one consumer. If a second test
+ * file ever needs it, promote it to its own file under tests/Fakes/.
+ */
+class SpyIdBrokerClient extends FakeIdBrokerClient
```
briskt: It is very rare to include a second class in a PHP file. I agree that it does seem to belong here, and I would be fine with leaving it here for now. One consideration, though, is that it may be forgotten and another identical or similar class may be added later rather than utilizing this one.

briskt (approved):
Much better! I love the idea of a spy/fake for this.

modules/mfa/tests/MfaTest.php
samuel-sowpati-sil: Agreed it's the right habit even for a single-consumer class. I went ahead and moved `SpyIdBrokerClient` to its own file (modules/mfa/tests/SpyIdBrokerClient.php) in the latest commit

PR #396, opened 2026-04-09
use docker healthcheck for test containers

briskt:
[ITSE-1081](https://support.gtis.sil.org/issue/ITSE-1081) where possible, replace whenavail with health checks

---

### Changed (non-breaking)
- Use docker healthcheck instead of [whenavail](https://github.com/sil-org/whenavail-script) for the test containers.


briskt:
The flaky integration test seems more persistent with this change. Closing until that is resolved.

PR #399, opened 2026-04-13
Add Behat step to verify wrong password does not update last_login

samuel-sowpati-sil:
[IDP-1807](https://support.gtis.sil.org/issue/IDP-1807) Consider adding tests to ssp-base for when to update last_login_utc
---

### Added
- `last-login.feature` — Behat scenario 1 (wrong password must not update `last_login_utc`).
- `LastLoginContext.php` — step definitions that query real broker via HTTP before and after a wrong-password login against `ssp-idp2.local`, then assert `last_login_utc` 
- `last_login_features` suite entry in `behat.yml`.

---

### DIFF
--- a/features/bootstrap/LoginContext.php
+++ b/features/bootstrap/LoginContext.php
@@ -170,22 +170,6 @@ class LoginContext extends FeatureContext
     }
 
     /**
-     * Verify that last_login_utc was NOT updated. In the silauth login path,
-     * updateUserLastLogin() only exists in the downstream MFA module — if
-     * authentication fails here, the MFA module never runs and last_login
-     * is never touched. (IDP-1807 scenario 1)
-     *
-     * @Then the last login should not have been updated
-     */
-    public function theLastLoginShouldNotHaveBeenUpdated()
-    {
-        Assert::false(
-            $this->authenticator->isAuthenticated(),
-            'Expected authentication to have failed (which prevents any last_login update).'
-        );
-    }
-
-    /**
      * @Given I do not provide a username
      */
     public function iDoNotProvideAUsername()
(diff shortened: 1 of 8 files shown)


features/bootstrap/LoginContext.php
```diff
@@ -169,6 +169,22 @@ function () use ($authenticator) {
...
+     * updateUserLastLogin() only exists in the downstream MFA module — if
+     * authentication fails here, the MFA module never runs and last_login
+     * is never touched. (IDP-1807 scenario 1)
+     *
+     * @Then the last login should not have been updated
+     */
+    public function theLastLoginShouldNotHaveBeenUpdated()
+    {
+        Assert::false(
+            $this->authenticator->isAuthenticated(),
```
briskt: I don't believe that the actual assertion should be so different than the phrase that defines it. This leads to a fragile state where any change in the assumption you stated in the comment will render this test invalid. Fortunately, that's very unlikely to change in this case. Given that fact, one would conclude that this test isn't very important. In addition, this same assertion is part of the "I should not be allowed through" step, so this adds nothing to the scenario.

This scenario probably can only be verified using a full integration test as we talked about [earlier](https://github.com/sil-org/ssp-base/pull/395#discussion_r3031162079).

samuel-sowpati-sil: Thanks Steve — that's a fair catch. The assertion body not matching the step phrase was exactly the kind of fragility that makes a test worse than no test. I'll replace it with a real integration test that queries the broker directly for `last_login_utc` before and after a wrong-password attempt against ssp-idp2.local.

briskt (approved):
Nice!

PR #401, opened 2026-04-15
pass the username to the forgot password page on the profile manager

briskt:
[IDP-1944](https://support.gtis.sil.org/issue/IDP-1944) Pre-fill Forgot Password

---

### Changed (non-breaking)
- Changed the Forgot Password link to a button with onclick handler to add a query parameter containing the username.


hobbitronics (approved):
Looks good

PR #402, opened 2026-04-15
fix flaky tests

briskt:
[IDP-1925](https://support.gtis.sil.org/issue/IDP-1925) ssp-base test failure: Don't prompt for MFA

---

### Fixed
- Fixed flaky tests.
  - Don't retain the session in the context. Get a fresh copy as needed.
  - Use `wait()` to ensure the browser has finished redirecting to the destination.
  - Use a `wait()` on body content match instead of an instantaneous snapshot of the page content, just in case it is still changing.
  - Removed `submitSecondarySspFormIfPresent` as outdated for SimpleSAMLphp 2.x.
- Remove unused `use` lines.

### DIFF
--- a/features/bootstrap/ExpiryContext.php
+++ b/features/bootstrap/ExpiryContext.php
@@ -1,7 +1,6 @@
 <?php
 
 use Behat\Mink\Element\DocumentElement;
-use Behat\Mink\Session;
 use PHPUnit\Framework\Assert;
 
 /**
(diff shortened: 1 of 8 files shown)


features/bootstrap/ExpiryContext.php
mtompset: I don't think this is used anymore.

mtompset (approved):
By changing every thing to getSession() it grabs the mink session set in the construct method of FeatureContext which was modified to use a locally created session, as a property is not needed.

briskt:
> By changing every thing to getSession() it grabs the mink session set in the construct method of FeatureContext which was modified to use a locally created session, as a property is not needed.

Are you simply describing the solution presented in the PR or are you asking a question?

features/bootstrap/ExpiryContext.php
briskt: Good find.

PR #403, opened 2026-04-16
replace whenavail with Docker healthcheck

briskt:
[ITSE-1081](https://support.gtis.sil.org/issue/ITSE-1081) where possible, replace whenavail with health checks

---

### Changed (non-breaking)
- Replace [whenavail](https://github.com/sil-org/whenavail-script) with Docker [healthcheck](https://docs.docker.com/reference/compose-file/services/#healthcheck)

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -6,6 +6,14 @@ x-db_healthcheck: &db_healthcheck
     timeout: 5s
     retries: 3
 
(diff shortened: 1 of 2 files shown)


mtompset (approved):
The reuse syntax is a tad complicated for the novice reader.

PR #404, opened 2026-04-16
update to simpleSAMLphp version 2.5.0

briskt:
### Fixed
- Update SimpleSAMLphp to version 2.5.0, which requires:
  - Require PHP version 8.3 or later.
  - Install and require php-bcmath extension.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,11 +10,12 @@
     }
   ],
   "require": {
-    "php": ">=8.1",
+    "php": ">=8.3",
+    "ext-bcmath": "*",
     "ext-gmp": "*",
     "ext-json": "*",
     "codemix/yii2-streamlog": "^1.3",
-    "simplesamlphp/simplesamlphp": ">=2.2.2 <2.5.0",
+    "simplesamlphp/simplesamlphp": "~2.5.0",
     "simplesamlphp/composer-module-installer": "^1.0",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -10,11 +10,12 @@
     }
   ],
   "require": {
-    "php": ">=8.1",
+    "php": ">=8.3",
```
mtompset, line 13: We're only bumping to 8.3 now?! I thought we got all the projects.

mtompset (approved):

composer.json
briskt: :man_shrugging:

devon-sil:
@briskt With 2.5, I believe the login_hint patch is no longer necessary.  https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-2.5/modules/saml/src/IdP/SAML2.php

https://support.gtis.sil.org/issue/IDP-1927

briskt:
> @briskt With 2.5, I believe the login_hint patch is no longer necessary. https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-2.5/modules/saml/src/IdP/SAML2.php
> 
> https://support.gtis.sil.org/issue/IDP-1927

Thanks. I tried to find that ticket.

PR #407, opened 2026-04-29
Update simplewebauthn

briskt:
### Changed (non-breaking)
- Update simplewebauthn to the latest version.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ssp-base",
   "dependencies": {
-    "@simplewebauthn/browser": "^4.1.0"
+    "@simplewebauthn/browser": "^13.3.0"
```
mtompset, line 4: Wow! What a jump.

mtompset (approved):

package.json
briskt: Yeah, a lot can happen in 5 years. Most of the major version increments were for the server side, though.

PR #410, opened 2026-05-20
Disable admin module

briskt:
[IDP-112](https://support.gtis.sil.org/issue/IDP-112) Consider turning off `core:AdminPassword` authsource

---

### Changed (non-breaking)
- Disabled the admin module as a security precaution. It can be enabled by adding a `extra-config.php` file using an image layer or volume.

Example extra-config.php file:

```php
<?php
return [
    'module.enable' => [
        'admin' => true,
    ]
];
```

Example docker run command:

```
docker run github.io/sil-org/ssp-base -v ${PWD}/extra-config.php:/data/extra-config.php
```

Example Dockerfile:

```Dockerfile
FROM github.io/sil-org/ssp-base

COPY extra-config.php /data/extra-config.php
```

### DIFF
--- a/dockerbuild/config/config.php
+++ b/dockerbuild/config/config.php
@@ -655,7 +655,7 @@ $config = [
 
     'module.enable' => [
         'core' => true,
-        'admin' => true,
+        'admin' => false,
         'saml' => true,
         'expirychecker' => true,
         'material' => true,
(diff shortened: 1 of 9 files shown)


briskt:
This disables the admin module rather than removing the AdminPassword auth source. I _think_ that's the only place the admin password could be used, though tomorrow I may look at removing the auth source as well.

dockerbuild/config/config.php
```diff
@@ -1484,3 +1484,12 @@
     $config['authproc.idp'][48] = 'sildisco:TagGroup';
     $config['authproc.idp'][49] = 'sildisco:AddIdp2NameId';
 }
+
+$extraConfigFile = '/data/extra-config.php';
+
+if (file_exists($extraConfigFile)) {
+    $extraConfig = include $extraConfigFile;
+    if (is_array($extraConfig)) {
```
hobbitronics, line 1492: Why would it be an array?

hobbitronics (approved):
Good idea to turn it off by default

dockerbuild/config/config.php
briskt: It looks like this:

```
[
    'module.enable' => [
        'admin' => true,
        'exampleauth' => true,
    ]
]
```

That's the way it's designed so it's easy to pass to array_replace_recursive.

PR #411, opened 2026-05-21
update PHPUnit to 10.5

briskt:
[IDP-1149](https://support.gtis.sil.org/issue/IDP-1149) update PHPUnit in ssp-base

---

### Fixed
- Updated PHPUnit to 10.5.

### Changed
- Removed the metadata test that runs against the SP container. It is broken while there's a different version of PHPUnit in play (test vs released ssp-base) and there's not much benefit to test SP metadata anyway.

### DIFF
--- a/.github/workflows/test-and-publish.yml
+++ b/.github/workflows/test-and-publish.yml
@@ -14,6 +14,10 @@ jobs:
         run: cp local.env.dist local.env && cp local.broker.env.dist local.broker.env
       - name: test
         run: docker compose run test ./run-tests.sh
+      - name: check hub metadata for tests
+        run: docker compose run ssp-hub.local ./run-metadata-tests.sh
+      - name: check idp metadata for tests
+        run: docker compose run ssp-idp1.local ./run-metadata-tests.sh
 
   build-and-publish:
     name: Build and Publish Docker image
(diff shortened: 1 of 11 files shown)


devon-sil:
@briskt Are the "extraneous metadata tests" in the test containers testing different IdP configurations?

briskt:
> @briskt Are the "extraneous metadata tests" in the test containers testing different IdP configurations?

I suppose you could say that. They're not very realistic configurations, though. If we want to keep them, they'll have to be disabled until this change is released. There's a circular dependency that breaks the sp one at least, since it uses a released version of ssp-base that isn't compatible between PHPUnit versions. I'll try it with just the idp tests.

briskt:
> > @briskt Are the "extraneous metadata tests" in the test containers testing different IdP configurations?
> 
> I suppose you could say that. They're not very realistic configurations, though. If we want to keep them, they'll have to be disabled until this change is released. There's a circular dependency that breaks the sp one at least, since it uses a released version of ssp-base that isn't compatible between PHPUnit versions. I'll try it with just the idp tests.

Actually, I was thinking the run-tests.sh script also ran the metadata tests against the hub, but it runs on the "test" container. Not sure how much good that does. Anyway, it's just the metadata test on the sp1 container that is broken, and since we don't use ssp-base for any real SPs, that one can probably be removed permanently.

hobbitronics (approved):

devon-sil:
@briskt Thanks for looking through the tests more to make sure we aren't eliminating someting we needed.

PR #415, opened 2026-05-22
Mark metadata tests Incomplete (not Skipped) when SPs use SkipTests

samuel-sowpati-sil:
[IDP-2034](https://support.gtis.sil.org/issue/IDP-2034) CI tests are skipping more that just ones flagged for 'SkipTests'

---
### Changed (non-breaking)
- `MetadataTest.php` — three tests now call `markTestIncomplete` (yellow `I`) instead of `markTestSkipped` (yellow `S`) when SPs have `SkipTests`; the incomplete message lists the skipped SPs.

---

### DIFF
--- a/tests/MetadataTest.php
+++ b/tests/MetadataTest.php
@@ -336,9 +336,11 @@ class MetadataTest extends TestCase
         $spEntries = $metadata->getList('saml20-sp-remote');
 
         $badSps = [];
+        $skippedSps = [];
 
         foreach ($spEntries as $spEntityId => $spEntry) {
             if (!empty($spEntry[self::SkipTestsKey])) {
+                $skippedSps[] = $spEntityId;
                 continue;
             }
 
@@ -351,6 +353,11 @@ class MetadataTest extends TestCase
         $this->assertEmpty($badSps,
             'At least one SP has saml20.sign.response set to false ... ' .
             var_export($badSps, True));
+
+        if ($skippedSps) {
+            fwrite(STDERR, "Skipped saml20.sign.response check for SPs with the '" . self::SkipTestsKey .
+                "' flag: " . var_export($skippedSps, True) . "\n");
+        }
     }
 
     public function testMetadataSignAssertion()
@@ -360,9 +367,11 @@ class MetadataTest extends TestCase
         $spEntries = $metadata->getList('saml20-sp-remote');
 
         $badSps = [];
+        $skippedSps = [];
 
         foreach ($spEntries as $spEntityId => $spEntry) {
             if (!empty($spEntry[self::SkipTestsKey])) {
+                $skippedSps[] = $spEntityId;
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


tests/MetadataTest.php
```diff
@@ -353,11 +351,6 @@ public function testMetadataSignResponse()
         $this->assertEmpty($badSps,
             'At least one SP has saml20.sign.response set to false ... ' .
             var_export($badSps, True));
-
-        if ($skippedSps) {
-            $this->markTestSkipped('At least one SP had the ' . self::SkipTestsKey .
```
briskt, line 358: Is the `markTestSkipped` causing the `assertEmpty` to be ignored? I don't think it works that way. What are the mechanics that cause the `assertEmpty` to be ignored? 

As a separate, minor issue, I'm thinking we may want to continue reporting the list of skipped SPs. Would it work to simply change `markTestSkipped` to `markTestIncomplete`?

samuel-sowpati-sil: @briskt Good catches on both — and you're right about the mechanics.

- On the "ignored assertion": I overstated it in the description. `assertEmpty` was never ignored — it ran first every time. The trailing `markTestSkipped` only relabeled a passing run as Skipped (yellow), because the hub always has flagged SPs. So it could mask a green pass, never a failure. I've corrected the PR description.

- Regarding `markTestIncomplete`: it throws and halts the method the same way `markTestSkipped` does, so it'd just turn the yellow `S` into a yellow `I` — still not a green pass. To keep the test green and have visibility, I switched the block to print the skipped SP list instead. The assertion still gates pass/fail, the test reports green, and the skipped SPs show in the run output. 

- Side note: `testMetadataCerts` still skips silently (no printed list). I left it untouched to keep this PR scoped — happy to add the same print there as a follow-up if we'd like the audit trail to be consistent across all four.

briskt: OK, so if I understand correctly, there isn't really a bug to fix since the `SkipTests` feature is working as intended. It merely labels the test results incorrectly. Correct? I suppose you could say that mislabeling test results is a bug, but it doesn't change the overall pass/fail result of the test suite.

I still think `markTestIncomplete` is the closest annotation to the actual behavior. The test is only truly complete if no SkipTest flags were set. I asked Gemini about this, and it indicated that writing directly to stderr is "hacky". The alternatives are `markTestSkipped`, `markTestIncomplete`, and [TestDox](https://docs.phpunit.de/en/12.5/testdox.html). While TestDox is intriguing, it's probably out of scope for this issue.

briskt (approved):

tests/MetadataTest.php
briskt: Forgot to comment on `testMetadatCerts`. I appreciate that you kept the scope narrow, but yes, I think that while this is fresh on your mind, it would be good to refine that test also -- in a separate PR.

samuel-sowpati-sil: Thanks for the clarification. You're right it was mislabeling (I altered the PR title to match what's been done here). I've updated the PR description to reflect the new commit's changes. Will open a new PR for `testMetadataCerts` once this gets approved. Also going to post a clarification on ITSE-2744 / IDP-2034 — the original "no SP gets checked" framing was wrong.

briskt (approved):

PR #425, opened 2026-06-08
Refactor password field layout and visibility toggle

ethancanne:
[IDP-2014](https://support.gtis.sil.org/issue/IDP-2014) Request: show/hide option on password field
---

### Changed (non-breaking)
- Fixed the password visibility toggle display bugs by adding side-by-side layout instead of overlapping using position absolute
- 
Screenshot 2026-06-09 at 2 00 44 PM

### DIFF
--- a/modules/material/public/styles.2.3.6.css
+++ b/modules/material/public/styles.2.3.6.css
@@ -271,29 +271,31 @@ a.mdl-button.mdl-typography--body-2, a.mdl-button.mdl-typography--body-2 > i.mat
 
 
 /* Password field show/hide toggle */
+.login-inputs-wrapper {
+  display: flex;
+  flex-direction: column;
+  align-items: stretch;
+  width: 100%;
+  max-width: 300px;
+}
+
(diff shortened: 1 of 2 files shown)


briskt (approved):
It would be very helpful to have one or more screenshots on a PR that changes the user interface, especially when it's largely a styling change.

modules/material/public/styles.2.3.6.css
```diff
@@ -271,29 +271,31 @@ a.mdl-button.mdl-typography--body-2, a.mdl-button.mdl-typography--body-2 > i.mat
...
-  width: 32px !important;
-  height: 32px !important;
-  min-width: 32px !important;
-  padding: 0 !important;
-}
-
-.password-visibility-toggle .material-icons {
-  font-size: 20px;
-  line-height: 32px;
+  flex-shrink: 0;
```
hobbitronics, line 298: Nice css code smell cleanup.

hobbitronics (approved):

